PR #1, opened 2015-02-04
Tickets/dm 215

fritzm:


src/Interval.h
```diff
@@ -0,0 +1,301 @@
...
+/// `Interval` represents a closed interval of the real numbers
+/// by its upper and lower bounds. It is parameterized by the Scalar
+/// type, which must be constructible from a double.
+///
+/// An interval with identical upper and lower bounds contains
+/// a single point, and is equal to that scalar bound.
+///
+/// An interval with an upper bound less than its lower bound is
+/// empty (contains no points), as are intervals with NaN as
+/// either bound.
```
TallJimbo: Does Interval behave appropriately when either or both bounds are infinite?  If not, it might be worth calling that out explicitly in the documentation.

src/Interval.h
```diff
@@ -0,0 +1,301 @@
...
+        return (_a == x && _b == x) || (x != x && isEmpty());
+    }
+    bool operator!=(Scalar x) const { return !(*this == x); }
+
+    ///@{
+    /// `a` returns the lower endpoint of this interval. The return value
+    /// for empty intervals is arbitrary.
+    Scalar a() const { return _a; }
+    Scalar & a() { return _a; }
+    ///@}
```
TallJimbo: These accessors (and the ones below for `b`, `center`, and `size`) don't conform to our coding standards, which require `getA`, `getB`, `getCenter`, `getSize`.

Also, if you have an explicit need for a non-const reference accessor, fine - but I think a setter is preferable if not, or perhaps even just leaving the class immutable aside from the assignment operator.  I've frequently found that for simple classes it can turn out to be convenient for all mutation to go through the constructor, especially if the class is ever exposed to Python.

src/Interval.h
```diff
@@ -0,0 +1,301 @@
...
+
+    /// `isEmpty` returns true if this interval does not contain any points.
+    bool isEmpty() const {
+        return !(_a <= _b); // returns true when _a and/or _b is NaN
+    }
+
+    /// `center` returns the center of this interval. It is arbitrary
+    /// for empty intervals.
+    Scalar center() const { return 0.5 * (_a + _b); }
```
TallJimbo: Should add a newline here.

src/Interval.h
```diff
@@ -0,0 +1,301 @@
...
+    }
+    bool contains(Interval const & x) const {
+        if (x.isEmpty()) {
+            return true;
+        } else if (isEmpty()) {
+            return false;
+        }
+        return _a <= x._a && _b >= x._b;
+    }
+    ///@}
```
TallJimbo: I'm not convinced that it's a good idea for the empty interval to be contained by all other intervals, especially if it's also disjoint from all other intervals, and interval boundaries are always closed.  I'm pretty sure I implemented the `afw::geom::Box` containment methods to do the opposite: an empty box is never contained by another box.

src/Interval.h
```diff
@@ -0,0 +1,301 @@
...
+
+    ///@{
+    /// `shrunkTo` returns the intersection of this interval and x.
+    Derived shrunkTo(Scalar x) const { return Interval(*this).shrinkTo(x); }
+    Derived shrunkTo(Interval const & x) const {
+        return Interval(*this).shrinkTo(x);
+    }
+    ///@}
```
TallJimbo: I like these `shrinkTo` and `shrunkTo` names for intersection operations!  I'm also obligated to point out that the latter doesn't conform to our naming standards, but I think this case merits an exception (of course, I'm not sure if I'm empowered to grant them).

TallJimbo: Oh, and I guess I should add that `afw::geom::Box` uses `clipTo` for in-place intersection (but doesn't have a version that returns a new object).  It might be good to make that consistent, and go with `clipTo` and `clippedTo` here.

src/Angle.h
```diff
@@ -0,0 +1,105 @@
...
+    Angle & operator*=(double a) { *this = *this * a; return *this; }
+    Angle & operator/=(double a) { *this = *this / a; return *this; }
+
+    /// `degrees` returns the value of this angle in units of degrees.
+    double degrees() const { return _rad * DEG_PER_RAD; }
+    /// `radians` returns the value of this angle in units of radians.
+    double radians() const { return _rad; }
+
+    /// `isNormalized` returns true if this angle lies in the range [0, 2π).
+    bool isNormalized() const { return _rad >= 0.0 && _rad <= 2.0 * PI; }
```
TallJimbo: It probably makes sense to have a method to normalize the angle as well as test its normalization status.

src/AngleInterval.h
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    AngleInterval() : Base() {}
+
+    explicit AngleInterval(Angle x) : Base(x) {}
+
+    AngleInterval(Angle x, Angle y) : Base(x, y) {}
+
+    AngleInterval(Base const & base) : Base(base) {}
+};
```
TallJimbo: I think this class essentially requires the `Angle`s its constructed with to be normalized (or be un-normalized in a consistent way, if that makes any sense) in order for any the topological operations it inherits from `Interval` to work.  And even if the inputs are normalized, I don't think it necessarily handles intervals that cross the wrapping point correctly.

To me, that suggests that `AngleInterval` should normalize its inputs on construction (selectively, of course - the upper bound should never be normalized to a value below the lower bound), and users should not be allowed to modify its internals directly (which would of course mean you couldn't allow that for other `Interval`s either, if you want to preserve the inheritance).

src/Angle.cc
```diff
@@ -0,0 +1,41 @@
...
+#include <ostream>
+
+namespace lsst {
+namespace sg {
+
+std::ostream & operator<<(std::ostream & os, Angle const & a) {
+    char buf[32];
+    std::snprintf(buf, sizeof(buf), "%.17g rad", a.radians());
+    return os << buf;
+}
```
TallJimbo: I know no one really likes iostream's precision manipulation tools, but that still seems better to me than using c-string operations into a temporary buffer just before you dump it into an iostream anyway.

src/AngleInterval.h
TallJimbo: Just saw `NormalizedAngleInterval`.  Sorry for the noise.

src/Region.h
```diff
@@ -0,0 +1,120 @@
...
+/// is that r1.relate(r2) will end up calling R2::relate(R1 const &). In other
+/// words, the call is polymorphic in the types of both r1 and r2.
+///
+/// One negative consequence of this design is that one cannot implement new
+/// Region types outside of this library.
+class Region {
+public:
+    virtual ~Region() {}
+    /// `clone` returns a deep copy of this region.
+    virtual Region * clone() const = 0;
```
TallJimbo: M

src/SpatialRelation.h
```diff
@@ -0,0 +1,59 @@
...
+
+/// Given a bitfield describing the spatial relations between two regions
+/// A and B (i.e. the output of `A.relate(B)`), `invertSpatialRelations`
+/// returns the bitfield describing the relations between B and A
+/// (`B.relate(A)`).
+inline int invertSpatialRelations(int relations) {
+	return (relations & (DISJOINT | INTERSECTS)) |
+	       ((relations & CONTAINS) << 3) |
+	       ((relations & WITHIN) >> 3);
+}
```
TallJimbo: This may be marginally more efficient, but I think it'd be a lot more readable if written out as an `if` or `switch` block rather than via bitshifting.

src/Angle.h
TallJimbo: As with `NormalizedAngleInterval`, I wrote the above before I saw `NormalizedAngle`; sorry for the noise.

TallJimbo:
A couple of general comments so far:
 - While we still have a lot of these lines in old code, I'm not a big fan of putting Doxygen `\author` tags in files, as they're obsoleted as soon as someone else edits the file, and we should be using git to track that anyway.
 - To conform to our coding standards, your classes generally need a lot more whitespace - newlines between function declarations.  I think exceptions can be made (even if it's not in the standards) for groups of very closely-related functions (e.g. operator overloads) that each only occupy one line, but I think anything that merits its own Doxygen comment definitely merits a newline to separate it from its neighbors.

src/Orientation.h
```diff
@@ -0,0 +1,123 @@
...
+/// This is equivalent to computing the scalar triple product a · (b x c),
+/// which is the sign of the determinant of the 3x3 matrix with a, b
+/// and c as columns/rows.
+///
+/// The implementation proceeds by first computing a double precision
+/// approximation, and then falling back to arbitrary precision arithmetic
+/// when necessary. Consequently, the result is exact.
+inline int orientation(UnitVector3d const & a,
+                       UnitVector3d const & b,
+                       UnitVector3d const & c)
```
TallJimbo: This implementation belongs in the .cc file.

src/Interval.h
smonkewitz: It should behave properly.

src/Interval.h
smonkewitz: Reading the code for afw::geom::Box2I:

    bool geom::Box2I::contains(Box2I const & other) const {
        return other.isEmpty() ||
            (all(other.getMin().ge(this->getMin())) && all(other.getMax().le(this->getMax())));
    }

it looks like the empty afw::geom::Box2I is also contained in all other boxes. I think this choice makes sense given the simple set-theoretic definition of containment I'm using.

src/Interval.h
smonkewitz: I forgot about this whole function names should start with a verb thing. That's going to pop up all over the place. I'll also try and get rid of the non-const reference accessors. I'm not sure I want to ditch the in place shrink/expand etc... method though - do you foresee those causing problems in Python land as well?

src/Interval.h
smonkewitz: Will do for clip. I will also take advantage of the SHOULD (rather than MUST) in the coding standards to keep the past tense verb names around. Those names feel clear and concise, and I would be sad to replace them with something like getShrunkenCopyOf.

src/Angle.cc
smonkewitz: This may be old school and a little clunky, but I far prefer it to writing a bunch of code that saves the iostream state, sets precision, unsets the floating point format (to get the equivalent of %g), calls a stream operator, then restores state. If you feel very strongly about this, I will reconsider.

src/Region.h
```diff
@@ -0,0 +1,120 @@
...
+/// Given a partitioning of the unit sphere with partitions that can be
+/// bounded by Regions, this provides all the necessary functionality for
+/// determining which partitions may intersect a Region.
+///
+/// When implementing a new concrete region subclass R, the Region interface
+/// should be extended with:
+///
+///     virtual int relate(R const &) const = 0;
+///
+/// All other other Region subclasses must then implement this method. In
```
fritzm: typo "other other"

src/Region.h
smonkewitz: So, I'm not a fan of classes making memory management decisions for their clients. To me, returning a shared_ptr from a method says - "here, have a pointer to this thing I am also retaining a pointer to", which is not at all what's going on. Unfortunately, it seems that auto_ptr has issues, and no other pre-C++11 construct is a good fit. If C++11 were OK (in headers, not just source files), you could talk me into std::unique_ptr without much trouble.

Is shared_ptr<Region>(r->clone()) so terrible?

src/Angle.cc
TallJimbo: I think your feelings on this are stronger than mine.

src/Interval.h
TallJimbo: I think the value of the in-place shrink/expand outweighs any problems that might appear in Python later, which I think can be at least partially worked around by just not exposing those methods to Python.

I should be a bit less cryptic on this point, though: I think composition in Python just works much better with immutable objects (this is much of why Python has immutable strings and tuples, I think), because it also tends to share everything.  That's not always a good idea in C++ of course, but you can also get some of the benefits by wrapping a mutable C++ class into something that's immutable from the Python side.  For instance, that lets C++ code return a contained object by const reference without worrying that a Python user could directly modify it.

src/Interval.h
TallJimbo: I stand corrected.  Every time I encounter this question, I can't decide what the right behavior should be.  If you're confident that this one is more useful than the other, that's fine with me.

src/Interval.h
TallJimbo: That resolution is fine with me.

src/Region.h
TallJimbo: I think the proscription against C++11 in headers is irrelevant given that the motiviation behind that is just to ensure we don't confuse Swig (and there's no Swig here).  Like so many other issues (I'll try to put together a list at the end of the revue), you should get sign off on this from K-T, but I'd say you should just use `unique_ptr` here.

IMO, `shared_ptr(r->clone())` isn't so terrible on its own, because there's no way anyone would write a `clone()` method that returns an internal reference, but as soon as you have any other methods that return raw pointers, it's just too easy to assume the wrong thing.

TallJimbo:
One place I do have a concern about rolling-your-own instead of using an existing third-party library is in the custom matrix/vector classes, where it seems like you could have used Eigen (with perhaps a custom Eigen-based object for UnitVector3d).  Unlike Boost, Eigen is header-only and hence trivial to install, and while you'd only be using a tiny fraction of Eigen's functionality here, you've written a fair amount of very repetitive code that would be unnecessary if you were to depend on Eigen.

src/Chunker.h
```diff
@@ -0,0 +1,157 @@
...
+
+#ifndef LSST_SG_CHUNKER_H_
+#define LSST_SG_CHUNKER_H_
+
+/// \file
+/// \author Serge Monkewitz
+/// \brief This file declares a class for partitioning the sky into chunks
+///        and sub-chunks.
+
+#include <stdint.h>
```
TallJimbo: I think it's considered better practice to include `<cstdint>` and use e.g. `std::int32_t` below.

src/Chunker.h
```diff
@@ -0,0 +1,157 @@
...
+
+    Angle getOverlap() const {
+        return _overlap;
+    }
+    int32_t getNumStripes() const {
+        return _numStripes;
+    }
+    int32_t getNumSubStripesPerStripe() const {
+        return _numSubStripesPerStripe;
+    }
```
TallJimbo: The above methods could use some documentation (especially `getOverlap()`, as I think I could guess what the others do).

src/Chunker.h
smonkewitz: I agree, but unfortunately, I found out yesterday that <cstdint> is C++11, and switched to stdint.h so that builds on RHEL 6 work at all.

src/Circle.h
```diff
@@ -0,0 +1,260 @@
...
+/// representation that uses the center vector and cosine of the circle
+/// opening angle, remains accurate for circles with very small opening
+/// angles.
+class Circle : public Region {
+public:
+    static Circle empty() { return Circle(); }
+    static Circle full() { return Circle(UnitVector3d::Z(), 4.0); }
+
+    static double squaredChordLengthFor(Angle openingAngle);
+    static Angle openingAngleFor(double squaredChordLength);
```
TallJimbo: Need documentation for these.

src/Circle.h
```diff
@@ -0,0 +1,260 @@
...
+        if (isFull() || x.isEmpty()) {
+            return true;
+        }
+        if (isEmpty() || x.isFull()) {
+            return false;
+        }
+        NormalizedAngle cc(_center, x._center);
+        return _openingAngle >
+               cc + x._openingAngle + 4.0 * Angle(MAX_ASIN_ERROR);
+    }
```
TallJimbo: This implementation should go in the .cc file.

src/Circle.h
```diff
@@ -0,0 +1,260 @@
...
+            return true;
+        }
+        if (isFull() || x.isFull()) {
+            return false;
+        }
+        NormalizedAngle cc(_center, x._center);
+        return cc > _openingAngle + x._openingAngle +
+                    4.0 * Angle(MAX_ASIN_ERROR);
+    }
+    ///@}
```
TallJimbo: Also belongs in .cc file.

src/NormalizedAngle.h
```diff
@@ -0,0 +1,164 @@
...
+        NormalizedAngle x;
+        double c = 0.5 * (a.radians() + b.radians());
+        if (a <= b) {
+            x._a = Angle(c);
+        } else {
+            // The result is (a + b + 2π) / 2, normalized to [0, 2π)
+            x._a = Angle((c < PI) ? (c + PI) : (c - PI));
+        }
+        return x;
+    }
```
TallJimbo: This method and the one above it should go in the .cc file.

src/UnitVector3d.h
```diff
@@ -0,0 +1,204 @@
...
+    /// the given spherical coordinates.
+    UnitVector3d(Angle lon, Angle lat) {
+        double sinLon = sin(lon);
+        double cosLon = cos(lon);
+        double sinLat = sin(lat);
+        double cosLat = cos(lat);
+        _v.x() = cosLon * cosLat;
+        _v.y() = sinLon * cosLat;
+        _v.z() = sinLat;
+    }
```
TallJimbo: Should go in .cc.

src/Vector3d.h
```diff
@@ -0,0 +1,218 @@
...
+    }
+    bool operator<(Vector3d const & v) const {
+        if (_v[0] == v._v[0]) {
+            if (_v[1] == v._v[1]) {
+                return _v[2] < v._v[2];
+            }
+            return _v[1] < v._v[1];
+        }
+        return _v[0] < v._v[0];
+    }
```
TallJimbo: Should go in .cc file - if you even need it.  For objects that don't have an obvious sort order, I tend to lean towards not providing a sorting operator on the class, and passing a comparison functor as a template argument if you need it for a particular container.

src/Vector3d.h
```diff
@@ -0,0 +1,218 @@
...
+            scale = 3.7291703656001034e-155;
+            invScale = 2.6815615859885194e+154;
+            n2 = ((*this) * scale).squaredNorm();
+        }
+        double norm = std::sqrt(n2);
+        _v[0] = (_v[0] * scale) / norm;
+        _v[1] = (_v[1] * scale) / norm;
+        _v[2] = (_v[2] * scale) / norm;
+        return norm * invScale;
+    }
```
TallJimbo: Should go in .cc file.

I also wonder how much you really want to "properly" handle normalization of truly tiny vectors; I would guess that most of the time a vector as tiny as the ones you're worried about here would have actually been zero in exact arithmetic, and throwing an exception is just as appropriate - if not ore appropriate - than dealing with denormalized floating point values.

src/Interval.h
smonkewitz: I'm honestly not sure what's more useful in these edge cases - I went with the simple definition because it made it easier for me to figure out how the code should behave.

TallJimbo:
I think you should probably go with the usual LSST directory structure.  I don't think that would require adding a dependency on sconsUtils or anything; it's really not difficult to avoid sconsUtils even if you have separate include/ and lib/ directories for the headers and output library.

TallJimbo:
Is the only usage of `BigInteger` in the implementation of `orientationExact`?  If so, I'd like to get a better understanding of how necessary that is - all of your topological relationship operations already have the potential to just return "indeterminate", so naively it seems like being limited to a less precise, floating-point-only `orientation` implementation would simply result in more indeterminate results.  Is that correct?

I don't feel like I'm qualified to judge whether having our own arbitrary-precision arithmetic library is a good idea vs. using an external package, but if it's actually optional to some degree, I think that makes depending optionally on a third-party library a more attractive option.

TallJimbo:
Returning to the general question of the usefulness of immutability, I think it might be worth considering whether it'd be a good idea to do that for all `Region` subclasses.  Essentially, that's because using `shared_ptr`s on immutable classes is much less problematic than it is on mutable classes, and we tend to use `shared_ptr` most frequently on large and/or polymorphic objects, and wrapping to Python effectively uses shared pointers on everything.

Of course, if you expect minor modifications to large `Region`s frequently, you don't have much choice but to add mutators.  But if you don't have a specific need for mutators, I'd leave them off until you know they're needed.

smonkewitz:
Regarding the LSST directory structure - I thought I was producing it on install, but actually my install scons target is broken (will fix). At one point, it was convenient for Daniel to have all the code in one directory, but that may no longer be the case. I'll double check.

smonkewitz:
The only usage of `BigInteger` is indeed in `orientationExact`, and `orientationExact` is the only exact operation required if one wants to implement error free point in polygon and polygon-polygon relation tests. I think your naive interpretation is more or less correct.

That said, there are a few reasons I'd like to keep this approach:

- Figuring out how indeterminate things are can be hard (notice I basically punted when it came to ellipses)
- The library isn't very useful if things are too indeterminate.
- In my opinion, being exact makes the implementation of some algorithms (e.g. convex hull) much more understandable.
- I ran out of time, but I'd like to implement support for non-convex polygons (so we can represent e.g. more complicated observational footprints). I don't know how to implement point-in-simple-polygon on the sphere except by looking for polygon edges that cross a geodesic between the point being tested and a point known to be in the polygon (i.e. some vertex). I'm glossing over the details a bit, but basically if you get the count wrong, which can happen with a non-exact orientation test if say your geodesic just grazes the polygon, you get incorrect containment results for points that are wildly outside the polygon in question.

smonkewitz:
Regarding region immutability - I think I can do this for the more complicated region types. Circle and Box are a bit special though, because they are used to bound more complex things (it's kind of nice to be able to grow a Box to encompass all the vertices in a polygon). I suppose I could get rid of all the `clipTo`/`expandTo` mutators, leaving just the `clippedTo`/`expandedTo` functions and hope the optimizer is good at copy elision.

src/Vector3d.h
smonkewitz: Fair point. I'll introduce a functor in the convex polygon implementation.

src/UnitVector3d.cc
```diff
@@ -0,0 +1,85 @@
...
+        n.x() = 1.0;
+    }
+    return UnitVector3d(v.cross(n));
+}
+
+UnitVector3d UnitVector3d::orthogonalTo(Vector3d const & v1,
+                                        Vector3d const & v2)
+{
+    Vector3d n = (v2 + v1).cross(v2 - v1);
+    if (n == Vector3d()) {
```
fritzm: How about having a Vector3d::isZero() predicate to use here instead?  It might be more readable, and at least conceptually avoid redundant construction of zero vectors in these comparisons?

src/UnitVector3d.cc
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+        n.x() = 1.0;
+    }
+    return UnitVector3d(v.cross(n));
+}
+
+UnitVector3d UnitVector3d::orthogonalTo(Vector3d const & v1,
+                                        Vector3d const & v2)
+{
+    Vector3d n = (v2 + v1).cross(v2 - v1);
```
fritzm: Would a Vector3d::robustCross be generally useful enough to have here instead of coding inline?

src/UnitVector3d.cc
smonkewitz: Great idea - will do.

src/Circle.h
```diff
@@ -0,0 +1,251 @@
...
+
+    /// `squaredChordLengthFor` computes and returns the squared chord length
+    /// between points in S² that are separated by the given angle. The squared
+    /// chord length l² and angle θ are related by l² = 4 sin²(θ/2).
+    static double squaredChordLengthFor(Angle openingAngle);
+
+    /// `openingAngleFor` computes and returns the angular separation between
+    /// points in S² that are separated by the given squared chord length. The
+    /// squared chord length l² and angle θ are related by l² = 4 sin²(θ/2).
+    static Angle openingAngleFor(double squaredChordLength);
```
fritzm: It seems to me the returned angle here would be implicitly normalized -- was there a reason not to return the more strongly constrained NormalizedAngle here?

smonkewitz: At the moment, you can pass a negative opening angle or squared chord length to these two functions to get a squared chord length / angle corresponding to an empty circle, so the returned value isn't necessarily normalized. If you're OK with that design, I'll fix the documentation to suck less.

fritzm: Oh okay -- seems like a common pattern throughout the library, so I'm okay with it.

One overall consideration with that I think is whether these sorts of degenerate parameters are *only* ever used in construction of empty boxes/angles/intervals/etc (in which case having some sort of explicit empty constructors might be better) or if its an expectation that degenerate parameters might pop up in e.g. intermediate results of the various algos and then be sensibly and correctly consumed by other things down the line (which makes the code more simple and robust so then I'm all for it.)

smonkewitz: In this case, it does simplify the implementation of e.g. `Circle::dilateBy()`, so I'm inclined to leave it. But it wouldn't be hard to change it and I don't feel particularly strongly either way. I'm also still on the fence about giving range restricted values their own classes to be honest.

PR #2, opened 2015-12-09
Tickets/dm 2262

smonkewitz:
This ticket converts sphgeom to a stack-like package, i.e. a package with standard layout and which uses sconsUtils to build. Note that it is still missing a python interface - that will come later.


.gitignore
```diff
@@ -8,5 +12,21 @@ doc/doxygen.conf
 doc/html
 doc/xml
 doc/*.tag
-include
-lib
+python/lsst/sphgeom/version.py
+tests/.tests/
```
jbecla: should tests/.tests go into git?

ups/sphgeom.table
```diff
@@ -1,2 +1,7 @@
+setupRequired(base)
+setupRequired(sconsUtils)
+setupOptional(doxygen)
+
 envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
 envPrepend(DYLD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(PYTHONPATH, ${PRODUCT_DIR}/python)
```
timj: This also needs LSST_LIBRARY_PATH to be set.

smonkewitz: Not objecting, but why? What is LSST_LIBRARY_PATH?

Also, should I make the same change in ups/TMPL.table in the templates package?

timj: I need it for any stack package to work on OS X. See http://dmtn-001.lsst.io for more details. All other packages have received this change.

smonkewitz: Thanks for the link. Note however that the templates package did not receive this update (maybe it's not the primary source of documentation for this sort of thing, but I didn't know where else to look).

timj: Ah. I did not realise you were talking about the templates repo. Sorry. Right, I had updated that with the `bin.src` change but not the table file. I'll do that.

timj: Template updated: https://github.com/lsst/templates/commit/72b8fcd0cbdbce965b4a23413307c1300d03d943

PR #3, opened 2016-02-12
Document Ellipse constructor

fjammes:


src/Ellipse.cc
```diff
@@ -96,6 +96,14 @@ Ellipse::Ellipse(UnitVector3d const & f1, UnitVector3d const & f2, Angle alpha)
...
 }
 
+/*
+ * @brief Create an ellipse representation
+ *
+ * @param center: ellipse center (ra, decl)
+ * @param alpha: size of semi-major-axis angle in radian
+ * @param beta: size of semi-minor-axis angle in radian
+ * @param orientation: ellipse orientation, in radian
+ */
```
andy-slac: Do we really need it here, there is already a documentation for all constructors in header file?

fjammes: Yes, I've missed the header file, this can be removed. Thanks

PR #7, opened 2016-08-22
DM-7315: Update tests for pytest

timj:


mtpatter:
@AstroVPK, it looks like there are some unused imports that can be taken out- sys is imported but not used in all 8 test*.py files and lsst.sphgeom.WITHIN is imported but not used in testAngleIntervals.py and testInterval1d.py.

Also, since the executable line `#!/usr/bin/env python` wasn't there and seems to introduce problems with OSX, I'd rather default to not adding it in this merge.

PR #8, opened 2016-08-29
DM-5533: Add HTM indexing to sphgeom

smonkewitz:


andy-slac:
Serge, in my code I implemented this method to print HTM index values:

    def _htm_repr(index, level):
        """
        Human-friendly representation of HTM index.
        """
        res = ''
        while level >= 0:
            res = ('0', '1', '2', '3')[index % 4] + res
            level -= 1
            index >>= 2
        res = {2: 'S', 3: 'N'}.get(index, 'X') + res
        return res

Could something like this be added to C++ HTM pixelization class (and for other types of indices if it makes sense)?

Also I thought that defining explicit `typedef` for index type instead of using `uint64_t` everywhere could make it more readable, but that is not a big deal.

doc/htm.md
```diff
@@ -0,0 +1,37 @@
+HTM Indexing
+============
+
+Overview                        {#htm-overview}
+========
+
+The Hierarchical Triangular Mesh (HTM) is described in the work of of A. Szalay,
```
andy-slac: `of of`

smonkewitz:
Sure, but before I add it - can I ask why you want this? Is it just so you have some idea of where a point is based on its index? For Q3C and modified-Q3C, I'm not sure how to best stringify the cube face number - does starting off the string with X, Y, Z, -X, -Y, -Z (i.e. the normal vector of the cube face) make sense to you?

doc/htm.md
```diff
@@ -0,0 +1,37 @@
+HTM Indexing
+============
+
+Overview                        {#htm-overview}
+========
```
andy-slac: This section title uses the same markdown level as document title above, would it be better to give it different level (e.g. ----------)?

andy-slac:
Yes, it's mostly for debugging like dumping the list of ranges, raw numbers look to unreadable. If there is no well-defined representation for cube indices then whatever you think is best is fine with me. Maybe explicit "+" in front of XYZ to make it more symmetrical with -XYZ.

doc/htm.md
smonkewitz: I'll try, but IIRC I had trouble getting doxygen to do sensible things with this (those weird braced anchors after the header are necessary in order to be able to link to sections from within code comments). Another thing I should point out it that the top-level README.md is also used by doxygen as the front-page for the docs it generates. The cost is that one of those braced anchors is visible when github renders the markdown, which I thought was acceptable. Feel free to disagree :)

doc/q3c.md
```diff
@@ -0,0 +1,121 @@
...
+similar scheme described below. They are provided by the
+[Mq3cPixelization](\ref lsst::sphgeom::Mq3cPixelization) class.
+
+To obtain the index of unit vector P, both the original and modified
+schemes start by centrally projecting P onto the faces of the cube [-1,1]³.
+Each of the 6 cube faces is assigned a distinct face number, and a local
+coordinate system (u, v). A grid of pixels is overlaid on each cube face,
+and pixels are labeled according to a space filling curve. The index is
+obtained by combining the label of the pixel containing the face coordinates
+with the face number.
```
andy-slac: Just for my education - how is an ambiguity between cube faces is resolved?

smonkewitz: If you mean, "given a point exactly on a cube edge or vertex, which face is it assigned to", I've left it as implementation defined, on the grounds that whatever choice one makes is an infinitesimal error away from a different choice. And even if I tried to match say what the Postgres extension module does (in theory), the code for computing indexes is different so that there are likely to be points very close to a cube face edge for which my implementation and the original one disagree.

include/lsst/sphgeom/constants.h
```diff
@@ -33,24 +33,25 @@ namespace sphgeom {
 // Note: given a compiler that does correctly rounded decimal to
 // binary floating point conversions, PI = 0x1.921fb54442d18p1 in
 // IEEE double precision format. This is less than π.
-static double const PI = 3.1415926535897932384626433832795;
-static double const RAD_PER_DEG = 0.0174532925199432957692369076849;
-static double const DEG_PER_RAD = 57.2957795130823208767981548141;
+static constexpr double PI = 3.1415926535897932384626433832795;
```
andy-slac: Is `static` useful here?

smonkewitz: As far as I know, `static` for a namespace scope variable specifies internal linkage and is orthogonal to `constexpr`. It's probably not necessary, but I'm not 100% sure.

smonkewitz:
Human-readable string conversions for pixel indexes have been added in 302a5fe.

doc/htm.md
smonkewitz: So yes, using `----` for a section breaks doxygen output if a label is present ([see this stackoverflow post](http://stackoverflow.com/questions/13648124/header-label-in-doxygen-markdown-page-makes-header-title-disappear)).

src/HtmPixelization.cc
```diff
@@ -0,0 +1,261 @@
...
+#include "PixelFinder.h"
+
+
+namespace lsst {
+namespace sphgeom {
+
+namespace {
+
+// Raw HTM root vertex coordinates. UnitVector3d objects are avoided so that
+// library clients cannot run into the static initialization order fiasco.
```
andy-slac: I'm not sure there could be a fiasco, `UnitVector3d` seems to be a trivial type.

src/Mq3cPixelization.cc
```diff
@@ -0,0 +1,346 @@
...
+// chosen such that the 8 or 9 pixel neighborhood of the center pixel is
+// guaranteed to contain the bounding circle. The minimum angular pixel width
+// could be precomputed per level. Alternatively, there is some constant W
+// such that the minimum angle between two points separated by at least one
+// pixel is greater than W * 2^-L at level L. Given the bounding circle
+// radius R, the subdivision level L of the initial neighborhood is the binary
+// exponent of W/R (and can be extracted by calling std::frexp).
+//
+// Finding W and implementing the above is left as a future optimization.
+template <typename RegionType, bool InteriorOnly>
```
andy-slac: I wonder why `InteriorOnly` has to be a part of the type? Does it do some smart template specialization based on its value?

src/PixelFinder.h
```diff
@@ -0,0 +1,176 @@
...
+/// \brief This file provides a base class for pixel finders.
+
+#include "lsst/sphgeom/RangeSet.h"
+
+#include "ConvexPolygonImpl.h"
+
+
+namespace lsst {
+namespace sphgeom {
+namespace {
```
andy-slac: I'd probably use `detail` namespace instead of anonymous here.

src/HtmPixelization.cc
smonkewitz: It does not have a trivial default constructor, so it is not a trivial type. Still, it does seem like the compiler should be able to avoid the fiasco. But I find it very hard to figure out what is actually supposed to happen, so I erred on the side of caution here.

src/Mq3cPixelization.cc
smonkewitz: See `src/PixelFinder.h` - making it part of the type means some of the code in `visit` and `_insert` can be evaluated (or eliminated) at compiler time rather than at run-time.

src/PixelFinder.h
smonkewitz: Why?

include/lsst/sphgeom/constants.h
smonkewitz: However: I did some more reading and it turns out that (unbeknownst to me) using const or constexpr implies internal linkage. (Even if this weren't the case, an unnamed namespace would be the more modern way to express internal linkage). So you are right to question, and no it isn't useful - I'll remove it. Thanks for catching this!

doc/htm.md
andy-slac: OK, let's keep it as it is, no big deal.

doc/q3c.md
andy-slac: Yes, I meant something on the edge or a corner of a cube, but I guess the same question is applies to the edges of pixels. Implementation-defined worries me. We are going to persist indices in the database, if implementation changes over lifetime of our data and ambiguity is resolved differently it may potentially cause some visible effects depending on what we do with indices. I'm not sure I can think of a specific example now but I'm still worried :worried:

src/HtmPixelization.cc
andy-slac: By trivial I meant that its constructor should not depend on other classes, so it supposed to be safe in a context of global initialization.

src/Mq3cPixelization.cc
andy-slac: OK, if it is an important optimization I'm fine with that.

src/PixelFinder.h
andy-slac: Mainly because anonymous namespace in a header file looks strange :smile: I'm not entirely sure what the purpose of that could be. Anonymous namespace becomes unique in every .cc file that includes this header, which also make classes contained in that namespace unique.  The `PixelFinder` class in file A.cc will be totally unrelated to the `PixelFinder` class in B.cc (even if they have all the same template parameters), and this could be confusing (why do you want to have multiple "copies" of the same class?)

smonkewitz:
Good point about making the stringification routines part of the `Pixelization` interface.

src/HtmPixelization.cc
smonkewitz: Sorry, I still don't understand. I'm imagining the case where you have a client that performs some sphgeom activity as part of the initialization activities for a static object - wouldn't they have to worry about when the constructors for sphgeom statics are called relative to their object, regardless of what the sphgeom constructors actually do?

If the `UnitVector3d` type were trivial (in the formal sense), or had an appropriate `constexpr` constructor, then it feels like the compiler/linker should be able to construct static const objects at compile time and stuff their representations into e.g. the .rodata  section of an executable/library, which would avoid the fiasco.

But I don't know what is guaranteed to happen and what isn't - perhaps my Google-fu is weak - so I stuck to primitive types. Am I missing or misunderstanding something?

doc/q3c.md
smonkewitz: I guess I could write down the current behavior, but my point is that this shouldn't assuage  your worries. The current implementation depends on libm, and the (ra, dec) -> pixel index calculation involves calls to trigonometric functions. Unless you include a correctly rounded (and bug-free) implementation of sin/cos with this library (IEEE does not mandate correct rounding except for +,-,*,/,sqrt), it seems to me that there is the potential for indexes to change simply by recompiling/relinking. Insisting on bit-identical results no matter what also means you wouldn't be able to change the order of operations or the operations themselves for optimization purposes, ever. So, things like using fused-multiply-add intrinsics on newer processors would be out.

For the record, the bottom-left edges and vertex of the square pixels in the Q3C pixelizations belong to the pixel. For HTM I tried to follow what an existing (non-public) HTM implementation does.

smonkewitz:
@andy-slac : I think i've addressed all your comments - would you like to give it another quick look before I merge?

src/HtmPixelization.cc
andy-slac: Well, the answer to your question is that people should not instantiate global objects which do non-trivial stuff in their constructors like calling other complex libraries :smile_cat:

The problem is partially mitigated by shared libraries, before anyone can call any method in a shared library all its global objects have to be initialized. But if you do static linking then you are right, you either have to do static initialization or do crazy singleton-like stuff. In any case your solution is the best if you worry about clients doing nasty things. My unclear comment was that fiasco is avoided if everyone is not doing complex stuff in their globals constructors, your classes definitely qualify.

TallJimbo: If you did want to use `UnitVector3d` and guard against the fiasco, you could just define this as a static variable in a function and have the variable return a reference to it.  That'd make it guaranteed to be initialized on first use, though it might make it a little more verbose for callers to access.

smonkewitz: @TallJimbo : good idea. I think it's OK to pay a small performance cost for accessing the array in question, and your idea makes the code more readable, so I will go ahead and make this change.

andy-slac:
@smonkewitz - everything looks perfect, thanks for answering my concerns, please merge :octocat:

PR #9, opened 2016-09-05
DM-7311: Python 3

fred3m:


fred3m:
No level 1 conversions necessary

include/lsst/sphgeom/Pixelization.h
```diff
@@ -23,6 +23,11 @@
 #ifndef LSST_SPHGEOM_PIXELIZATION_H_
 #define LSST_SPHGEOM_PIXELIZATION_H_
 
+// PyInt_GetMax is not in the python 3 C API, so we define it for python >=3
+#if PY_MAJOR_VERSION >= 3
+#define PyInt_GetMax(x) 2<<31
+#endif
```
timj: @smonkewitz do you have an opinion on this? It seems a bit weird to use 32bits when even on Python 2 we have 64bit ints in 64bit mode.

timj: Actually, surely we can't use a Python API in this file can we? It would have to go in the `.i` file.

fred3m: It built on both Jenkins and my machine, so I think it is ok to keep in in this file but I can change it if that is more clear.

As for the code, PyInt_GetMax() is defined to be LONG_MAX from climits.h, which is 32 or greater (see https://docs.python.org/2/c-api/int.html and http://www.cplusplus.com/reference/climits/). I could just replace this with LONG_MAX, which should work regardless of the python version.

timj: Building or not is not the issue. Putting Python API definitions into a header file that is not part of a Python interface is the problem. This define should absolutely go in `pixelizations.i` because it's part of the Python interface and that is where the Python API call is being used.

It looks like the code in `pixelizations.i` could work with just 64-bit but I'd like @smonkewitz to comment before we merge as it might be that we should be tweaking the code there and not using `PyInt_GetMax` at all.

smonkewitz: First of all, I think this should be addressed in python/lsst/sphgeom/pixelizations.i - adding Python C-API stand-ins to our C++ library headers feels like pollution to me.

Secondly: I would recommend trying to replace [this line](https://github.com/lsst/sphgeom/blob/tickets/DM-7311/python/lsst/sphgeom/pixelizations.i#L44) with:

    if (p <= static_cast<unsigned long long>(std::numeric_limits<long>::max())) {

which should be equivalent (based on my reading of the docs for `PyInt_GetMax`).

smonkewitz:
This package does not depend on lsst::utils, so it currently isn't getting:

    // Allow Python2 to accept unicode strings
    %begin %{
    #define SWIG_PYTHON_2_UNICODE
    %}

from `p_lsstSwig.i` the way most other packages do. I don't believe sphgeom has any API that accepts `std::string` arguments, but it does in places produce them. @timj / @fred3m - is this a problem? I'm thinking maybe we should add those lines at the top of sphgeomLib.i, if only to avoid surprises in the future.

timj:
Good point. I don't have a problem with them being added to the `.i` file. I added them in the `log` package as that also does not have access to `p_lsstSwig.i`.

python/lsst/sphgeom/pixelizations.i
```diff
@@ -29,6 +29,11 @@
     #include "lsst/sphgeom/Q3cPixelization.h"
 %}
 
+// Allow Python2 to accept unicode strings
+%begin %{
+#define SWIG_PYTHON_2_UNICODE
+%}
```
timj: I think @smonkewitz wanted that in `sphGeomLib.i`.

python/lsst/sphgeom/pixelizations.i
```diff
@@ -41,7 +46,7 @@
             o = PyNumber_Multiply(tmp, tmp);
             Py_DECREF(tmp);
         } else {
-            if (p <= static_cast<unsigned long long>(PyInt_GetMax())) {
+            if (p <= static_cast<unsigned long long>(std::numeric_limits<long>::max())) {
                 o = PyInt_FromLong(static_cast<long>(p));
```
timj: Can this use the modern `PyLong_FromLong` variant?

python/lsst/sphgeom/pixelizations.i
smonkewitz: Yes please!

python/lsst/sphgeom/pixelizations.i
smonkewitz: I defer to you both - it seems like PyIntObject no longer exists in 3? I am OK with always using PyLongObject even on Python 2.

fred3m:
Current changes build in Jenkins.

python/lsst/sphgeom/pixelizations.i
```diff
@@ -41,8 +41,8 @@
             o = PyNumber_Multiply(tmp, tmp);
             Py_DECREF(tmp);
         } else {
-            if (p <= static_cast<unsigned long long>(PyInt_GetMax())) {
-                o = PyInt_FromLong(static_cast<long>(p));
+            if (p <= static_cast<unsigned long long>(std::numeric_limits<long>::max())) {
+                o = PyLong_FromLong(static_cast<long>(p));
```
smonkewitz: Actually, if we avoid `PyInt_FromLong` even on Python 2, you can simplify this function to:

    inline PyObject * _rangeBeginningOrEndToIntOrLong(uint64_t p, bool isEndPoint) {
        PyObject * o;
        if (p == 0 && isEndPoint) {
            // Map trailing bookend to 2**64
            PyObject * tmp = PyLong_FromUnsignedLongLong(UINT64_C(0x100000000));
            if (!tmp) {
                return nullptr;
            }
            o = PyNumber_Multiply(tmp, tmp);
            Py_DECREF(tmp);
        } else {
            o = PyLong_FromUnsignedLongLong(p);
        }
        return o;
    }

smonkewitz:
LGTM (but @timj should confirm). Thanks for implementing all the suggested fixes!

timj:
Looks good to me as well. Just needs a bit of squashing. @smonkewitz thanks for the feedback.

fred3m:
No problem @smonkewitz, thanks for the help.

PR #11, opened 2017-02-21
Tickets/DM-9281 Wrap sphgeom with pybind11 instead of swig

r-owen:


include/lsst/sphgeom/Box.h
```diff
@@ -53,7 +53,7 @@ namespace sphgeom {
 /// - Box::allLongitudes().contains(b.getLon())
 class Box : public Region {
 public:
-    static constexpr uint8_t TYPE_CODE = 'b';
+    static uint8_t const TYPE_CODE;
```
pschella: This is sad, if it really can't work this way. `constexpr` really should be the way to go.

tests/testUnitVector3d.py
```diff
@@ -46,7 +51,7 @@ def testConstruction(self):
         self.assertAlmostEqual(u.x(), 0.5, places=15)
         self.assertAlmostEqual(u.y(), 0.5, places=15)
         self.assertAlmostEqual(u.z(), 0.5 * math.sqrt(2.0), places=15)
-        u = UnitVector3d.northFrom(u.asVector3d())
+        u = UnitVector3d.northFrom(u)
```
r-owen: This code would be clearer if used a new variable name each time instead of changing the value of `u` several times in the same method.

smonkewitz: OK.

python/lsst/sphgeom/angle.cc
```diff
@@ -0,0 +1,101 @@
...
+    cls.def(py::init<double>(), "radians"_a);
+    cls.def(py::init<Angle>(), "angle"_a);
+    // Construct an Angle from a NormalizedAngle, enabling implicit
+    // conversion from NormalizedAngle to Angle in python via
+    // py::implicitly_convertible
+    cls.def("__init__",
+            [](Angle &self, NormalizedAngle &a) {
+                new (&self) Angle(a.asRadians());
+            },
+            "normalizedAngle"_a);
```
smonkewitz: `NormalizedAngle` provides a cast operator for conversion to `Angle`, so this isn't needed on the C++ side. This was added on the Python side so that `py::implicitly_convertible` can work.

python/lsst/sphgeom/box3d.cc
```diff
@@ -0,0 +1,183 @@
...
+
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+namespace lsst {
+namespace sphgeom {
+namespace {
+
+// Convert a Python integer to a C++ int in [-3, 3). If the Python integer
+// is out of bounds, an IndexError is raised.
```
smonkewitz: You're right - I'll fix the comment and move the utility function into a header file.

python/lsst/sphgeom/ellipse.cc
```diff
@@ -0,0 +1,114 @@
...
+    cls.def("getAlpha", &Ellipse::getAlpha);
+    cls.def("getBeta", &Ellipse::getBeta);
+    cls.def("getGamma", &Ellipse::getGamma);
+    cls.def("complement", &Ellipse::complement);
+    cls.def("complemented", &Ellipse::complemented);
+
+    // Note that the Region interface has already been wrapped.
+
+    // The lambda is necessary for now; returning the unique pointer
+    // directly leads to incorrect results and crashes.
```
smonkewitz: I don't know if it's a bug or an oversight and I haven't had time to produce a reduced test-case (there is https://github.com/pybind/pybind11/issues/690, but it's filed against pybind11 2.0.1). It does feel like it should work though, and that returning a `std::unique_ptr` should probably be equivalent to a raw pointer return with `pybind11::return_value_policy:: take_ownership`.

python/lsst/sphgeom/ellipse.cc
```diff
@@ -0,0 +1,114 @@
...
+        return py::str("Ellipse({!s}, {!s}, {!s})")
+                .format(self.getF1(), self.getF2(), self.getAlpha());
+    });
+    cls.def("__repr__", [](Ellipse const &self) {
+        return py::str("Ellipse({!r}, {!r}, {!r})")
+                .format(self.getF1(), self.getF2(), self.getAlpha());
+    });
+    cls.def("__setstate__", [](Ellipse &self, py::bytes bytes) {
+        new (&self) Ellipse(*decode(bytes));
+    });
```
smonkewitz: Yes, both are for pickling. I implemented `__reduce__` wherever possible, since it is the LSST standard, and because using `__[gs]etstate__` instead of `__reduce__` means one is required to use `cPickle` instead of `pickle` in Python 2.7. Additionally, the pickling protocol must be set to 2, earlier versions don't work.

python/lsst/sphgeom/htmPixelization.cc
```diff
@@ -0,0 +1,67 @@
...
+
+    cls.def(py::init<int>(), "level"_a);
+    cls.def(py::init<HtmPixelization const &>(), "htmPixelization"_a);
+
+    cls.def("getLevel", &HtmPixelization::getLevel);
+
+    cls.def("__eq__",
+            [](HtmPixelization const &self, HtmPixelization const &other) {
+                return self.getLevel() == other.getLevel();
+            });
```
smonkewitz: Good catch, will do.

python/lsst/sphgeom/interval1d.cc
```diff
@@ -0,0 +1,67 @@
...
+
+    cls.def("isFull", &Interval1d::isFull);
+
+    cls.def("__str__", [](Interval1d const &self) {
+        return py::str("[{!s}, {!s}]").format(self.getA(), self.getB());
+    });
+    cls.def("__repr__", [](Interval1d const &self) {
+        return py::str("Interval1d({!r}, {!r})")
+                .format(self.getA(), self.getB());
+    });
```
smonkewitz: Yes it can, as can all the other interval-like classes you mention - see `include/lsst/sphgeom/interval.h`. There is no base class exposed to Python; instead, that header adds lambdas that call the various interval methods, including one for `__reduce__`. This works for the interval types with a CRTP base class, as well as `NormalizedAngleInterval`.

smonkewitz:
Russell - I think I've covered all your suggestions, except for getting rid of '///' for doxygen comments. I think this last point is out of scope - the doxygen comments would basically have to be rewritten to match the current LSST C++ style guide. Do you want to have another look before I merge?

r-owen:
I don't need another look. Your code already looked excellent.

PR #14, opened 2018-05-05
DM-14255: Add ConvexPolygon.intersects and related methods

r-owen:
Add the following methods to ConvexPolygon:
- contains(Region)
- isDisjointFrom(Region)
- intersects(Region)
- isWithin(Region)

### DIFF
--- a/src/ConvexPolygon.cc
+++ b/src/ConvexPolygon.cc
@@ -313,6 +313,22 @@ bool ConvexPolygon::contains(UnitVector3d const & v) const {
     return detail::contains(_vertices.begin(), _vertices.end(), v);
 }
 
+bool ConvexPolygon::contains(Region const & r) const {
+    return (relate(r) & CONTAINS) != 0;
+}
+
+bool ConvexPolygon::isDisjointFrom(Region const & r) const {
+    return (relate(r) & DISJOINT) != 0;
+}
+
+bool ConvexPolygon::intersects(Region const & r) const {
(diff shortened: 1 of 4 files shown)


src/ConvexPolygon.cc
```diff
@@ -313,6 +313,22 @@ bool ConvexPolygon::contains(UnitVector3d const & v) const {
     return detail::contains(_vertices.begin(), _vertices.end(), v);
 }
 
+bool ConvexPolygon::contains(Region const & r) const {
+    return (relate(r) & CONTAINS) != 0;
```
TallJimbo: I don't think the ` != 0` part of these tests is necessary.  Do you prefer them stylistically? (I don't but, not strongly.)

smonkewitz: I think it is - `std::bitset` does not have a conversion operator to a `bool`. Another option here is to write `(relate(r) & CONTAINS).any()` instead.

r-owen: Exactly. I am not fond of `!= 0` but it gets the job done. I am happy to switch to `bitset.any` if you like. I would have preferred to use `bitset.test` but there are no useful constants for that (for good reasons, as @TallJimbo pointed out in email).

TallJimbo: I suppose I might have a slight preference for `any()`, as it's a helpful reminder to the reader that these are not in fact integers, but it's not a strong preference.

PR #17, opened 2018-07-23
DM-15187: Modernize sphgeom pickle support for pybind11 2.

r-owen:


python/lsst/sphgeom/box.cc
```diff
@@ -161,9 +162,9 @@ PYBIND11_MODULE(box, mod) {
     cls.def("__repr__", [](Box const &self) {
         return py::str("Box({!r}, {!r})").format(self.getLon(), self.getLat());
     });
-    cls.def("__setstate__", [](Box &self, py::bytes bytes) {
-        new (&self) Box(*decode(bytes));
-    });
+    cls.def(py::pickle(
+            [](const Box &self) { return python::encode(self); },
+            [](py::bytes bytes) { return decode(bytes).release(); }));
```
pschella: Are you sure the explicit release is needed?

python/lsst/sphgeom/unitVector3d.cc
```diff
@@ -117,16 +117,14 @@ PYBIND11_MODULE(unitVector3d, mod) {
...
-        if (t.size() != 3) {
-            throw std::runtime_error("Invalid state!");
-        }
-        new (&self) UnitVector3d(UnitVector3d::fromNormalized(
-                t[0].cast<double>(), t[1].cast<double>(), t[2].cast<double>()));
-    });
+    cls.def(py::pickle([](UnitVector3d const &self) { return py::make_tuple(self.x(), self.y(), self.z()); },
+                       [](py::tuple t) {
+                           if (t.size() != 3) {
+                               throw std::runtime_error("Invalid state!");
```
pschella: You may want to be a bit more explicit.

python/lsst/sphgeom/unitVector3d.cc
```diff
@@ -117,16 +117,14 @@ PYBIND11_MODULE(unitVector3d, mod) {
...
+                           }
+                           return new UnitVector3d(UnitVector3d::fromNormalized(
+                                   t[0].cast<double>(), t[1].cast<double>(), t[2].cast<double>()));
```
pschella: I know this was already there, but why a heap one?

r-owen: I'm not sure what you are asking. Are you asking why this code returns a pointer instead of a vector? If so, my guess is because `UnitVector3d` has no move constructor -- though I don't know if the compiler would still manage to move the data in this situation.

python/lsst/sphgeom/box.cc
r-owen: Good question, but I was hoping you could tell me! I copied the code from the existing wrappers for `decode` (e.g. in `box.cc`, `ellipse.cc`, etc.). However, I did try building without `.release()` in the pickle wrapper (but not in the decode wrapper) and I didn't see any difference -- the unit tests passed. I figured it was safer to go with what was there, but it might be code that was never needed, or was only needed for older versions of `pybind11`. What do you think?

r-owen: http://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html this implies that the release is not needed (and perhaps not wanted).

r-owen: Based on more tests I see the following:
- If I remove `.release()` from `clone` or `decode` in `region.cc` I get segfaults
- If I remove `.release()` from `decode` in `box.cc` I also get segfaults
- If I remove `.release()` from the pickle wrapper in `box.cc` I do not get segfaults

Based on that, and the fact that I don't understand why `release` is needed and the pickle docs say that the first function should look like `__getstate__`, which is all `decode` is, as far as I can tell, I propose to leave the `.release()` in everywhere. If you understand more deeply what is going on and think I can safely remove `.release()` from just the pickle support then I'm happy to do that.

pschella: I agree with the proposed solution. Will have to look into this later. Perhaps file a new mini ticket? The `unique_ptr -> shared_ptr` shouldn't matter.

PR #19, opened 2019-04-25
DM-19392: add numpy-vectorized contains methods

TallJimbo:


timj:
@TallJimbo I've rebased your branch so that it can do the travis checks I just added.

tests/test_Box.py
```diff
@@ -96,6 +101,17 @@ def test_relationships(self):
         r = b4.relate(b1)
         self.assertEqual(r, DISJOINT)
 
+    def test_vectorized_contains(self):
+        b = Box.fromDegrees(200, 10, 300, 20)
+        x = np.random.rand(5, 3)
+        y = np.random.rand(5, 3)
+        z = np.random.rand(5, 3)
+        c = b.contains(x, y, z)
```
jmeyers314, line 109: Should also test non-contiguous memory here (and similar below).  Something like `b.contains(x[::2], y[::2], z[::2])`.

jmeyers314 (approved):
Hey @TallJimbo.  I was just pining for vectorized `contains` in `sphgeom` earlier today, and then I found this PR.  So I'm reviewing in hopes that this can be merged.  

Only one minor in-line comment, but also, I'm curious what you would think about overloading a 2-argument version of `contains` with a vector of longitude and vector of latitude.  My use case is to mark the ras and decs contained in a ConvexPolygon.  It's not too bad to write 

```python
x = np.cos(ras)*np.cos(decs) 
y = np.sin(ras)*np.cos(decs)
z = np.sin(decs)
``` 
and then use the 3-argument methods here, but simply writing `convexPolygon.contains(ras, decs)` (radians assumed) also seems reasonable to me.

TallJimbo:
> So I'm reviewing in hopes that this can be merged.

Thanks for reminding me about this.  It looks like it's tied up with some other PRs that might involve backwards-incompatible API changes (already RFC'd, but maybe back before we had our current deprecation policy).  I'll take a look and see if I can find a subset of the ticket that's easy to merge (including this PR) and just do it, punting any problematic API changes to another ticket.

PR #25, opened 2020-07-08
DM-25867: Make package pip installable

timj:


python/lsst/__init__.py
```diff
@@ -1,3 +1,2 @@
 import pkgutil
-import lsstimport
```
TallJimbo, line 2: I was curious to see that this was still here, and maybe you already knew this, but `lsstimport` is apparently totally empty now, and can be removed everywhere.

I'm also curious to see that the `pkgutil` bit below that hasn't changed - did we revert using namespace packages at some point?  If so, we could probably go simplify the daf_butler mypy build.  If not, is this relic a time bomb - I thought namespace packages and the `pkgutil` snippet weren't compatible, so we might be getting by on the fact that the right one tends to appear first in `sys.path`.

setup.cfg
```diff
@@ -1,3 +1,36 @@
...
+[options]
+zip_safe = False
+package_dir=
+    =python
+packages=find:
+setup_requires =
+  pybind11 >=2.5.0
+  setuptools_cpp >=0.0.1
+install_requires =
+  pyyaml >=5.1
```
TallJimbo, line 28: Is this true?  I'd have thought that the functionality that requires yaml does no harm and cannot even be invoked unless pyyaml is installed.

setup.py
```diff
@@ -0,0 +1,50 @@
...
+    """Get the module part of the file name"""
+    file = os.path.split(path)[1]
+    root = os.path.splitext(file)[0]
+    if not root:
+        raise RuntimeError(f"Got {root} from {path} vis {file}")
+    return root
+
+
+# Very inefficient approach since this compiles the maing sphgeom
+# library code for every extension rather than building everything once
```
TallJimbo, line 37: FWIW, I don't think this is _as_ inefficient as it initially seems.  Or, rather, the best alternative isn't as much better as you might think (at least in general):

 - Concatenating all source into a single file, compiling a single object file, and linking that into a single .so might be the fastest approach for a package this small, because it means the compiler only processing the pybind11 headers once, but it doesn't scale to very large packages like afw, and it doesn't parallelize at all.

 - Compiling each source file into a separate object file, and then linking them into a single .so is what we recommend in new LSST code (and what we should switch to here, if not on this ticket); that enables much better warm rebuilds with scons (unchanged source files don't get recompiled) and much smaller total library sizes (-> better import times), but it does still require the compiler to process the entire set of pybind11 headers for every source file.

requirements.txt
```diff
@@ -0,0 +1,6 @@
+pybind11 >=2.5.0
```
TallJimbo, line 1: So this directly blocks us from being able to use the pip/conda version to build the rest of the pipelines, right?  It's a newer pybind11 than ndarray supports, and maybe some of our other code.

setup.cfg
TallJimbo: Ah, just looked at the imports - I think the above is only true if we put some import guards either in `__init__.py` or `_yaml.py`, which _might_ be worth doing.  Depends, I guess, on whether declaring it here makes sure pyyaml appears first in some dependency-ordering somewhere when both are present.

setup.py
TallJimbo: I misunderstood some of this in the comments above.  Yes, recompiling all of sphgeom into each of these modules is definitely inefficient.

smonkewitz: Is this something that could be ameliorated by reviving the pybind11-fat-wrapper branch?

timj: @TallJimbo did mention that sphgeom isn't doing pybind11 how we are meant to do it these days. It would help if it was a single `_sphgeom.so` that was being imported it.  I might take a look at that since building and running all the tests in a minute is much better than 15 minutes.  I'm not entirely sure how to get setuptools to build the files in `src/` once and then use them (or how to get it to build a proper shared library and link against it).

requirements.txt
timj: On the plus side this package must be compatible with the new pybind11. This was an arbitrary cut and paste from the pybind11 documentation. Since it worked with 2.5 and we know it builds with the stack version it's not a problem to change it.

setup.py
timj: I'll see if I can rebase that branch.

timj:
Good news. With #26 implemented the build and test now only takes 3.5 minutes instead of 15 minutes. @smonkewitz thanks for having done all the work already!

python/lsst/__init__.py
timj: I was the person that turned lsstimport into a stub when I realized we only needed it in one place (afw I think).  Yes, I've given up on the namespace packages -- too much still goes wrong if you remove it. We get by because we fill PYTHONPATH up with all sorts of different options for that directory and something usually wins. I think mypy was trying to tell us something so probably best to put it back.

setup.cfg
timj: I could lower the dependency version a bit but v5 added a bunch more loaders and I've have to do lots of import protection on them. Since we've had v5 for a while now there didn't seem to be any harm in requiring v5.

setup.py
timj: #26 has merged!

smonkewitz: Yay!

setup.cfg
TallJimbo: I was actually just suggesting that this package didn't need to depend on pyyaml at all, from the perspective that there is no functionality it provides that cannot be used without pyyaml, except for functionality that can only be invoked _through_ pyyaml.  But I'm not sure that's worth guarding the imports.

timj: The tests will fail if the Yaml package is not there and the imports inside the module are not protected. Deliberately not installing it seems to be more work than requiring it

timj: On second thoughts you probably should protect them. You are right that it’s only usable if the module is there and if it’s not there you never care.

PR #29, opened 2020-10-28
Demo for setuptools / pip-ready packaging

jonathansick:
The purpose of this experimental PR is to explore how the packaging for sphgeom might be modified so that third-party packages can consume it via `pip` (and possibly even be distributed via PyPI and conda-forge in the future). The related ticket is [DM-27354](https://jira.lsstcorp.org/browse/DM-27354).

The main change is that the build dependencies are installed using a `pyproject.toml` (using the PEP 517 mechanism). This replaces the use of a `requirements.txt` file for installing `setuptools_cpp` and other dependencies, and which isn't really a solution for a pip-installable package.

I also introduced a tox-based build to help isolate builds and tests from the development environment.

~~At this point, it seems that the extensions are compiled successfully. Unfortunately, the extensions aren't importable. I don't know why yet.~~ It works now that I'm setting `packages` and `package_dir` correctly for the last namespace package layout.


jonathansick:
Superseded by #30

PR #30, opened 2020-11-03
DM-27354: Enable pip installation using pyproject.toml / PEP 517

jonathansick:
This PR makes sphgeom pip installable by downstream packages without requiring a requirements.txt file to pre-populate the environment. The solution here relies on the pyproject.toml file to ensure that build dependencies are installed before setuptools is called to build and install the package. This relies on the new-ish PEP 517 mechanism for specifying the build requirements.

This work also removes the extra `setuptools_cpp` dependency as it didn't seem to add any value compared to pybind11's built-in setuptools integration.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,12 +9,12 @@ This is not a complete definition.
   interface.
 """
 
-from setuptools import setup
-from setuptools_cpp import ExtensionBuilder, Pybind11Extension
 import glob
 
 # Importing this automatically enables parallelized builds
 import numpy.distutils.ccompiler  # noqa: F401
+from setuptools import setup
+from pybind11.setup_helpers import Pybind11Extension, build_ext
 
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -45,8 +39,6 @@
 
 setup(
     version=version,
-    long_description=long_description,
```
timj, line 48: is long_description no longer something that is used then? I thought it was how you got the readme onto the PyPI page?

timj (approved):
Looks great. One question.

setup.py
jonathansick: @timj This is handled by the `file:` directive in the `setup.cfg` file for code-less import of the README into the package metadata.

timj: I don't see `file` in `setup.cfg`...

PR #31, opened 2021-04-16
Support Arm64

ctslater:
There was a broken code path for non-x86-64 architectures, so I fixed that and also added arm64 to the fast path.

### DIFF
--- a/include/lsst/sphgeom/codec.h
+++ b/include/lsst/sphgeom/codec.h
@@ -23,12 +23,20 @@
 #ifndef LSST_SPHGEOM_CODEC_H_
 #define LSST_SPHGEOM_CODEC_H_
 
+// Optimized path requires little endian arch and support for unaligned loads.
+#if defined(__x86_64__) or (defined(__aarch64__) and defined(__LITTLE_ENDIAN__))
+#define OPTIMIZED_LITTLE_ENDIAN
+#endif
+
+#ifdef NO_OPTIMIZED_PATHS
+#undef OPTIMIZED_LITTLE_ENDIAN
+#endif
+
 /// \file
 /// \brief This file contains simple helper functions for encoding and
 ///        decoding primitive types to/from byte strings.
 
 #include <vector>
-
 
 namespace lsst {
 namespace sphgeom {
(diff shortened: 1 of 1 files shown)


include/lsst/sphgeom/codec.h
```diff
@@ -36,32 +36,32 @@ namespace sphgeom {
 /// `encode` appends an IEEE double in little-endian byte order
 /// to the end of buffer.
 inline void encodeDouble(double item, std::vector<uint8_t> & buffer) {
-#if defined(__x86_64__)
-    // x86-64 is little endian.
+#if defined(__x86_64__) or defined(__aarch64__)
+    // x86-64 and Apple Arm64 are little endian.
```
smonkewitz, line 40: I don't think `__aarch64__` is Apple specific, and AArch64 is not guaranteed to be little-endian: https://developer.arm.com/documentation/102376/0100/Alignment-and-endianness. I'm unsure whether big-endian implementations are a real concern, but maybe there's a preprocessor check for little-endianness that could be added to the one for `__aarch64__` (or failing that, `__APPLE__`)?

Also, it might be good to provide a way to force the fallback path to get used (like NO_SIMD in other parts of the code), so it can be tested without exotic hardware or cross-compilation.

ctslater: Hi @smonkewitz, both points are good. I've now gone down the rabbit hole of trying to find which compilers on which architectures define a macro that would be useful here, and it looks like at least Apple defines `__LITTLE_ENDIAN__` on arm64, so that should handle the most relevant case today without being over-inclusive.

smonkewitz (commented):
If I still worked for LSST, I would approve this. Since I don't, I feel like someone who does should also take a look...

fritzm:
@smonkewitz your money is, in fact, still good here :-)  I approve this as well, if that is helpful.

TallJimbo (approved):
I've been vaguely following the GitHub posts and am happy to sign off if @smonkewitz is.

The conversation did make me think of https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html.  That's probably out of scope for this ticket, if a change in philosophy is even warranted - this is not an area of expertise for me, just something relevant I ran across at one point from someone pretty darn sensible.

smonkewitz:
I generally agree with the blog post, except that performance sometimes matters. I was thinking that it might matter for these functions (e.g. if a bunch of polygons ended up stored in a database column as binary strings). But I'm not sure that's actually happening, and maybe it's not that big a cost. (I mean, I obviously didn't check what assembly gets generated for the fallback, or I would have caught the brokenness - 🤦 ). So I'd agree it's worth considering whether we actually need the fast path at all in a separate PR.

ctslater:
Thanks for the panoply of reviews; merged to master.

timj:
@ctslater Did you mean to use a user branch for this?

ctslater:
I created a ticket branch after creating the PR, and since we had useful conversations here I didn't want to make a new PR just to switch to the ticket branch. Both branches point to the same commit.

timj:
Very clever. That explains why it thinks it merged even though technically this branch never did 😄

PR #32, opened 2021-10-20
[DM-32254] Use PixelizationABC as the base for Python pixelization classes

tgoldina:
https://jira.lsstcorp.org/browse/DM-32254

- Added lsst.sphgeom.PixelizationABC.
- Registered lsst.sphgeom.Pixelization as a subclass of lsst.sphgeom.PixelizationABC

### DIFF
--- a/python/lsst/sphgeom/pixelization_abc.py
+++ b/python/lsst/sphgeom/pixelization_abc.py
@@ -1,0 +1,141 @@
+# This file is part of sphgeom.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


timj (approved):
Thanks

python/lsst/sphgeom/pixelization_abc.py
```diff
@@ -0,0 +1,141 @@
...
+            Pixel index.
+
+        Returns
+        -------
+        `lsst.sphgeom.Region`
+            The spherical region corresponding to the pixel with index i
+
+        Raises
+        ------
+        `InvalidArgumentExceptiont`
```
erykoff, line 64: This should be `InvalidArgumentException` (no t at the end).

python/lsst/sphgeom/pixelization_abc.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        pass
+
+    @abc.abstractmethod
+    def index(self, v: UnitVector3d) -> int:
+        """ Compute the index of the pixel.
+
+        Parameters
+        ----------
+        v: `lsst.sphgeom.UnitVector3d)`
```
erykoff, line 75: Remove final extra parenthesis.

python/lsst/sphgeom/pixelization_abc.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Returns
+        -------
+        `lsst.sphgeom.Region`
+            The spherical region corresponding to the pixel with index i
+
+        Raises
+        ------
+        `InvalidArgumentExceptiont`
+            If i is not a valid pixel index.
```
timj, line 65: The style guide says this should read something like:

> Raised if ``i`` is not a valid pixel index.

PR #34, opened 2021-10-28
DM-23308: Add CMake support

fritzm:
Adds ability to build package with CMake in addition to eupspkg (allows to be included as submodule in other CMake projects, e.g. Qserv.)

### DIFF
--- a/.github/workflows/cmake.yaml
+++ b/.github/workflows/cmake.yaml
@@ -1,0 +1,31 @@
+name: cmake
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  cmake:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install Python prereqs
+        run: pip install pybind11[global]
+
+      - name: CMake configure
+        run: cmake -B ${{github.workspace}}/build
+
+      - name: CMake build
+        run: cmake --build ${{github.workspace}}/build --parallel `nproc`
+
+      - name: CMake test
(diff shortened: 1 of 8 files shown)


src/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+add_library(sphgeom SHARED)
+
+target_compile_features(sphgeom PRIVATE 
+    cxx_std_14
```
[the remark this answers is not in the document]
fritzm: Rolled to 17

fritzm:
@timj GHA add to do cmake build and run C++ tests -- please have a look?

.github/workflows/cmake.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: CMake configure
+        run: cmake -B ${{github.workspace}}/build
+
+      - name: CMake build
+        run: cmake --build ${{github.workspace}}/build --parallel `nproc`
+
+      - name: CMake test
+        working-directory: ${{github.workspace}}/build
+        run: ctest  --parallel `nproc`
```
timj, line 31: This test doesn't seem to do anything to test the python code. You are going to have to install pytest as I do in the other action and then run the tests to make sure it really does work from python.

fritzm: @timj Added python tests to the cmake GHA; please have another look?

timj (approved):
Thanks for adding the python tests -- this makes it less likely someone will break cmake. If you also install `pytest-xdist` you can add `-n 2` to the pytest to make it run in parallel so be a little faster.

fritzm:
Thanks @timj -- at less than 3s for the Python tests compared to the C++ build/link times, I'm inclined to not worry on it at this point. :-)

Let me know when the version tool lands, and I'll happily help out with the retrofitting of that into the cmake here and over in `log`.

PR #35, opened 2021-11-16
DM-32605: Add HealpixPixelization with healpy

erykoff:


fritzm:
Hmm, is this the first library feature we have here that is Python-side-only (and if so, do we care?)

erykoff:
I have discussed this with @TallJimbo and for butler usage -- which is the target use case -- then the python-side-only is fine, as long as it conforms to the pixelization_abc.  As far as I can see, the only use of `HtmPixelization` in c++ is in `lsst/partition`.  Maybe somebody else can speak to how much of a problem that might be looking forward.

fritzm:
As (what I believe we are) the principle customers of `lsst/partition`, the Qserv crew can certainly live without a C++ healpix pixelization at this time (and if we did find need of it later, I don't see any reason why we couldn't look for a suitable C++ implementation and propose porting the Python layer over to it at that time.)

I'm sorta more interested, in a musing, beard-stroking kind of way, in the question of whether this is a new direction for `sphgeom`, and if we should expect to see more Python-only things showing up in here now.

erykoff:
I think it was just a lot easier to get the healpix support in this way ... it took me a couple of afternoons, while the c++ integration of healpix has stymied many since DM-13483 was ticketed in February of 2018.

python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -0,0 +1,174 @@
...
+        if isinstance(region, (Box, Ellipse)):
+            _circle = region.getBoundingCircle()
+        elif isinstance(region, Circle):
+            _circle = region
+        elif isinstance(region, ConvexPolygon):
+            _poly = region
+        else:
+            raise ValueError("Invalid region.")
+
+        # Find all pixels at an arbirarily higher resolution
```
TallJimbo, line 142:
```suggestion
        # Find all pixels at an arbtrarily higher resolution
```

python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -0,0 +1,174 @@
...
+                                      nest=True)
+
+        return pixels
+
+    def __eq__(self, other):
+        if isinstance(other, HealpixPixelization):
+            return (self._level == other._level)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
TallJimbo, line 165: I think Python now takes care of `__ne__` for you if you implement `__eq__` (it didn't always, but I think this has been true for a long time).

python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -0,0 +1,174 @@
...
+
+import numpy as np
+import healpy as hp
+
+from ._sphgeom import RangeSet, Region, UnitVector3d
+from ._sphgeom import Box, Circle, ConvexPolygon, Ellipse
+from .pixelization_abc import PixelizationABC
+
+
+class HealpixPixelization:
```
TallJimbo, line 31:
```suggestion
class HealpixPixelization(PixelizationABC):
```
Since this is a pure-Python class, you can inherit from `PixelizationABC` the usual way.  It's only the C++ ones that need to use `register`.

python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -0,0 +1,174 @@
...
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return f"HealpixPixelization({self._level})"
+
+    def __reduce__(self):
+        return (self.__class__, (self._level, ))
+
+
+PixelizationABC.register(HealpixPixelization)
```
TallJimbo, line 174:
```suggestion
```
As per previous comment, just use regular inheritance.

TallJimbo (approved):
Looks good!  I've only checked that the mathematical methods and tests look plausible (they do) rather than correct, but I think that's enough here.

Only very minor comments.

PR #37, opened 2022-05-15
DM-34721: miscellaneous cleanups and union/intersection regions

TallJimbo:


include/lsst/sphgeom/codec.h
```diff
@@ -80,6 +80,44 @@ inline double decodeDouble(uint8_t const * buffer) {
...
+/// `encodeU64` appends an uint64 in little-endian byte order
+/// to the end of buffer.
+inline void encodeU64(std::uint64_t item, std::vector<uint8_t> & buffer) {
+#ifdef OPTIMIZED_LITTLE_ENDIAN
+    auto ptr = reinterpret_cast<uint8_t const *>(&item);
+    buffer.insert(buffer.end(), ptr, ptr + 8);
+#else
+    union { uint64_t u; double d; };
+    d = item;
+    buffer.push_back(static_cast<uint8_t>(u));
```
fritzm, line 92: You might consider a `resize()` here and then just using array subscripting assignments, to avoid the redundant bounds checking in eight successive `push_back()` calls.  OTOH, there are less than ten of them, so it's probably a <shrug> :-)

include/lsst/sphgeom/python/utils.h
```diff
@@ -65,6 +65,22 @@ std::unique_ptr<R> decode(pybind11::bytes bytes) {
     return R::decode(buffer, n);
 }
 
+/// Create a vector of Region (or Region-subclass) pointers by copying the
+/// regions from a sized Python iterable (e.g. S == py::tuple).
+///
+/// Note that the pybind11 built-in STL conversions don't work, because they we
```
fritzm, line 71: typo: "they we"

src/CompoundRegion.cc
```diff
@@ -0,0 +1,196 @@
...
+    auto r2 = getOperand(1).relate(rhs);
+    return
+        // Both operands must be disjoint with the given region for the union
+        // to be disjoint with it.
+        ((r1 & DISJOINT) & (r2 & DISJOINT))
+        // Both operands must be within the given region for the union to be
+        // within it.
+        | ((r1 & WITHIN) & (r2 & WITHIN))
+        // If either operand contains the given region, the union contains it.
+        | ((r1 & CONTAINS) | (r2 & CONTAINS));
```
fritzm, line 161: What about cases where neither r1 nor r2 individually contain rhs, but the union of r1 and r2 does (for example, two overlapping circles that together contain an ellipse not entirely contained by either one individually?)

src/CompoundRegion.cc
```diff
@@ -0,0 +1,196 @@
...
+    return
+        // Both operands must contain the given region for the intersection to
+        // contain it.
+        ((r1 & CONTAINS) & (r2 & CONTAINS))
+        // If either operand is disjoint with the given region, the
+        // intersection is disjoint with it.
+        | ((r1 & DISJOINT) | (r2 & DISJOINT))
+        // If either operand is within the given region, the intersection is
+        // within it.
+        | ((r1 & WITHIN) | (r2 & WITHIN));
```
fritzm, line 192: I think there are also cases where neither operand is individually within rhs, but the intersection of the operands is (consider two overlapping circles, and rhs an ellipse that surrounds the intersection but extends outside both of the circles)

fritzm (commented):
This looks like a nice feature to have.  Some questions in comments about whether all cases are properly covered in the `relate()` methods?

Additionally interesting: compound regions are the first possibly non-convex regions in this library.   At first I wondered if this might violate some assumptions elsewhere in the library (e.g. in various pixelization codes) but I didn't see any issues with this offhand...

src/CompoundRegion.cc
TallJimbo: You're absolute right about the behavior, but we're saved by the limited guarantees made by the `Region` interface on the behavior; from the docs on `Region::relate`:
```
    /// Said another way: if the CONTAINS, WITHIN or DISJOINT bit is set, then
    /// the corresponding spatial relationship between the two regions holds
    /// conclusively. If it is not set, the relationship may or may not
    /// hold.
    ///
    /// These semantics allow for conservative relationship computations. In
    /// particular, a Region may choose to implement `relate` by replacing
    /// itself and/or the argument with a simplified bounding region.
```
It'd be nice to do better, but I think we'd at least need way for all `Region` classes to be able to compute some common set of shapes they bound (rather than are bounded by), and that's a lot of hard math for still not being able to formally guarantee any more than we do.

src/CompoundRegion.cc
TallJimbo: (other thread on related problem for unions applies here, too)

include/lsst/sphgeom/codec.h
TallJimbo: I'm a little worried it could be an accidental pessimization, because the pattern is that lots of functions like this are going to be pushing into the same `buffer` vector, and I don't actually know if `resize` is going to (when the capacity isn't big enough) allocate _just_ the amount of extra space needed or do the kind of "allocate more than you did last time" thing that keeps `push_back` amortized constant-time.  Do you happen to know?

fritzm: Good point.  I think you could call `capacity()` before `resize()` to avoid the undesired behavior, but I have also just realized the the `resize()` is going to have to do a default init on all eight slots, so that will have a cost as well.   

In any case, for eight items per batch I think it's hard to argue convincingly that there would be much of a win or loss here in the end.

PR #38, opened 2022-05-23
DM-34914: Defer loading of healpy.

erykoff:


python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -143,17 +150,16 @@ def _interior_pixels_from_region(self, nside: int, region: Region):
         if _circle is not None:
             center = _circle.getCenter()
             vec = np.array([center.x(), center.y(), center.z()]).T
-            pixels = hp.query_disc(nside,
-                                   vec,
-                                   _circle.getOpeningAngle().asRadians(),
-                                   inclusive=False,
-                                   nest=True)
+            pixels = hp.query_disc(
```
timj, line 153: Why did this code get reorganized? (and the other places).

timj (approved):
Looks okay but I'd rather not have code reformatting in the same commit as the import change.

PR #41, opened 2022-06-03
Add meson build system for C++

bonimy:
I'm writing a project using meson as the build system as sphgeom
happens to be a dependence. There were a few issues trying to build with
cmake and I found it easier to just add meson support.


src/meson.build
```diff
@@ -0,0 +1,67 @@
...
+    'NormalizedAngle.cc',
+    'NormalizedAngleInterval.cc',
+    'orientation.cc',
+    'Q3cPixelization.cc',
+    'RangeSet.cc',
+    'Region.cc',
+    'UnitVector3d.cc',
+    'utils.cc',
+    'Vector3d.cc']
+sphgeom = static_library(
```
eli-schwartz, line 62: Why a static library, particularly? If you just use `library()` it becomes the choice of the builder (via `-Ddefault_library=static`, but it also accepts `=both` to build both at the same time).

The cmake build uses shared, not static, though cmake doesn't really have a good way to choose between the two.

meson.build
```diff
@@ -0,0 +1,9 @@
+project('sphgeom', 'cpp')
+
+subdir('src')
+
+pkg_mod = import('pkgconfig')
```
eli-schwartz, line 5: In addition to creating a pkg-config file, you may also want to create a `declare_dependency()` that adds the incdir and the sphgeom library, as an interface, which then allows using this meson.build as a Meson subproject with automatic dependency fallback.

src/meson.build
```diff
@@ -0,0 +1,67 @@
+incdir = include_directories('../include/')
+
+install_headers(
+    '../include/lsst/sphgeom/Angle.h',
+    '../include/lsst/sphgeom/AngleInterval.h',
```
eli-schwartz, line 5: Personally, I would define this in `subdir('include')` so you don't need all these ../ everywhere. :)

eli-schwartz (changes requested):

bonimy:
Just bumping to say that I've been in the process of moving states, but I'll hopefully finish this PR soon once things settle down.

timj:
I rebased this PR just in case there is still interest in it.

bonimy:
Sorry for never getting back to this. I ended up switching career paths.

If someone wants to take over this branch to merge it, please do. I haven't worked with meson for so long that I don't think I'd remember how to finish this branch out.

src/meson.build
bonimy: We can do just `library()`. I would assume that I didn't know I had the option at that time.

src/meson.build
bonimy: I didn't know that was possible. No objections from me.

PR #44, opened 2022-06-11
DM-3519: Add test matrix for python versions

mwittgen:


timj (approved):

mwittgen:
Will be done in DM-35347

PR #46, opened 2022-07-14
DM-35347: Switch to pyproject.toml

mwittgen:


pyproject.toml
```diff
@@ -1,12 +1,81 @@
 [build-system]
-build-backend = "setuptools.build_meta"
 requires = [
-    "setuptools >= 46.6.0",
+    "setuptools",
+    "lsst-versions >= 1.3.0",
     "wheel",
     "pybind11 >= 2.5.0",
     "numpy >= 1.18",
```
timj, line 7: Is this because we need the numpy C header files for the build with pybind11?

pyproject.toml
```diff
@@ -1,12 +1,81 @@
...
+license-files = ["LICENSE"]
+
+[tool.setuptools.dynamic]
+version = { attr = "lsst_versions.get_lsst_version" }
+
+[tool.black]
+line-length = 110
+target-version = ["py38"]
+
+[tool.isort]
```
timj, line 62: Oh, we hadn't run isort or black before. That explains the failures. Please fix the formatting problems.

timj (approved):
Looks good. Thanks for sorting out black/isort. Can you also please add a pre-commit configuration?

pyproject.toml
mwittgen: Yes.

mwittgen:
Reordered commits to have all black/isort changes in one. Added `.git-blame-ignore-revs`

PR #50, opened 2022-08-15
DM-35894: numpy.distutils requires pinning setuptools<65

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Build and install
         run: |
-          python -m pip install --upgrade "pip<22" setuptools
+          python -m pip install --upgrade pip "setuptools<65" "flake8<5"
           python -m pip install .[yaml,test]
 
       - name: Run tests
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install --upgrade setuptools wheel build
+          pip install --upgrade "setuptools<65" "flake8<5" wheel build
```
timj, line 57: Is this flake8 pin needed here? No tests are run.

.github/workflows/build.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install --upgrade setuptools wheel cibuildwheel
+          pip install --upgrade "setuptools<65" "flake8<5" wheel cibuildwheel
```
timj, line 101: Remove flake8 pin?

timj (approved):

.github/workflows/build.yaml
mwittgen: Without it fails as it run `pytest` in the workflow.

mwittgen: Sorry wrong line.

PR #52, opened 2022-12-13
DM-35692: unpin setuptools

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -22,15 +22,15 @@ jobs:
...
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
           cache-dependency-path: "setup.cfg"
 
       - name: Build and install
         run: |
-          python -m pip install --upgrade pip "setuptools<65" "flake8<5"
+          python -m pip install --upgrade pip setuptools "flake8<5"
```
timj, line 33: I think you need to also make similar changes to cmake.yaml. (the python@v4 and setuptools)

mwittgen: Updated `cmake.yaml` as well

timj (approved):

PR #65, opened 2024-05-08
DM-44261: Add support for parsing IVOA POS string

timj:
@gpdf here is a quick implementation for parsing IVOA POS strings.

* I copied the `continueClass` code from `lsst.utils` since I'm not sure we want to add utils as a dependency.
* I put all the code in `__init__.py` since the package is a bit oddly designed in terms of the code all coming from C++.
* There are no real tests that I did things properly since writing a test that the range/circle/polygon is correct is a bit tricky since if I have implemented this incorrectly I would mostly be creating a sphgeom Region in the same incorrect way in the test. Ideally I need a third party to write some tests.

cc/ @TallJimbo 

## Checklist

- [x] ran Jenkins


python/lsst/sphgeom/__init__.py
```diff
@@ -37,3 +38,130 @@
...
+        # class.
+        attr = cls.__dict__.get(name, None) or getattr(cls, name)
+        if _isAttributeSafeToTransfer(name, attr):
+            setattr(orig, name, attr)
+    return orig
+
+
+def _inf_to_lat(lat: float) -> float:
+    """Map +Inf to +90 and -Inf to -90 degrees."""
+    import math
```
andy-slac, line 84: Is this to keep `math` from module-level `vars`? I'm worried that there is non-zero price to pay for this on each call.

python/lsst/sphgeom/__init__.py
```diff
@@ -37,3 +38,130 @@
...
+        https://ivoa.net/documents/SIA/20151223/REC-SIA-2.0-20151223.html#toc12
+        for a description of the POS parameter but in summary the options are:
+
+        * ``CIRCLE <longitude> <latitude> <radius>``
+        * ``RANGE <longitude1> <longitude2> <latitude1> <latitude2>``
+        * ``POLYGON <longitude1> <latitude1> ... (at least 3 pairs)``
+
+        Units are degrees in all coordinates.
+        """
+        shape, *coordinates = pos.split()
```
andy-slac, line 123: Maybe do `strip` before `split` to protect from odd spaces?

python/lsst/sphgeom/__init__.py
```diff
@@ -37,3 +38,130 @@
...
+        elif shape == "RANGE":
+            import math
+
+            if n_floats != 4:
+                raise ValueError(f"RANGE requires 4 numbers but got {n_floats} in '{pos}'.")
+            # POS allows +Inf and -Inf in ranges. These are not allowed by
+            # Box and so must be converted.
+            # If either of the longitude values are infinite, all longitudes
+            # should be included.
+            if math.isinf(coordinates[0]) or math.isinf(coordinates[1]):
```
andy-slac, line 141: So even if both longitudes are the same sign `Inf` it still translates to 0-360? The standard does not say which combinations of `Inf` are allowed here, maybe something like this does not make sense. `RANGE -Inf 90 ...` seems to be converted to `RANGE 0 360 ...`? Standard also mentions that `NaN` is valid, but does not say what it means, and we do not handle it here?

python/lsst/sphgeom/__init__.py
```diff
@@ -37,3 +38,130 @@
...
+        coordinates = tuple(float(c) for c in coordinates)
+        n_floats = len(coordinates)
+        if shape == "CIRCLE":
+            if n_floats != 3:
+                raise ValueError(f"CIRCLE requires 3 numbers but got {n_floats} in '{pos}'.")
+            center = LonLat.fromDegrees(coordinates[0], coordinates[1])
+            radius = Angle.fromDegrees(coordinates[2])
+            return Circle(UnitVector3d(center), radius)
+        elif shape == "RANGE":
+            import math
```
andy-slac, line 133: `math` here too.

tests/test_ivoa.py
```diff
@@ -0,0 +1,66 @@
...
+            "POLYGON 12 34 14 34 14 36 12 36",
+            "RANGE 0 360 89 90",
+        )
+        for pos in example_pos:
+            region = Region.from_ivoa_pos(pos)
+            self.assertIsInstance(region, Region)
+
+        # Badly formed strings raising ValueError.
+        bad_pos = (
+            "circle 12 34 0.5",
```
andy-slac, line 54: Hmm, lower case is not good?

andy-slac (approved):
Looks good, few minor questions/comments.

python/lsst/sphgeom/__init__.py
timj: Ok. I will see about moving the entire content into its own file so that there won't be leakage into the caller namespace.

python/lsst/sphgeom/__init__.py
timj: Indeed. I was staring at the examples and trying to work out what the meant. Maybe `90 +Inf` is meant to mean `90 360`. With wraparound I was taking any Inf in longitude to mean the full range. Maybe @gpdf knows what they really meant. I also don't know what `NaN` is meant to do.

tests/test_ivoa.py
timj: I don't know. Maybe? Also something for @gpdf ...

python/lsst/sphgeom/__init__.py
timj: Doesn't look like I need to:
```python
>>> " a b  c        d       ".split()
['a', 'b', 'c', 'd']
```
so leading and trailing spaces are dropped anyhow.

python/lsst/sphgeom/__init__.py
andy-slac: You can add `del math` at the bottom like you do with `typing`.

timj: Yes, although for reasons unexplained I couldn't get math to work yesterday. I'm seeing if I can move it into a separate file.

python/lsst/sphgeom/__init__.py
andy-slac: Ah, OK, my memory is not working well, it's probably only useful if you specify `sep` character.

python/lsst/sphgeom/__init__.py
timj: James Dempsey has clarified that -Inf means 0 and +Inf means 360 so I've updated the code. He also stated that NaN is no longer part of the standard.

tests/test_ivoa.py
timj: James Dempsey confirms that `circle` is not allowed.

timj:
I have added `Region.to_ivoa_pos()` so you can now generate POS strings from a region.

timj:
@andy-slac I changed enough that it's probably worth a quick once over again. Thanks.

python/lsst/sphgeom/__init__.py
```diff
@@ -28,7 +28,9 @@
 
 """lsst.sphgeom
 """
+import typing
```
andy-slac, line 31: Is not needed anymore?

python/lsst/sphgeom/_continue_class.py
```diff
@@ -0,0 +1,218 @@
...
+        Units are degrees in all coordinates.
+        """
+        shape, *coordinates = pos.split()
+        coordinates = tuple(float(c) for c in coordinates)
+        n_floats = len(coordinates)
+        if shape == "CIRCLE":
+            if n_floats != 3:
+                raise ValueError(f"CIRCLE requires 3 numbers but got {n_floats} in '{pos}'.")
+            center = LonLat.fromDegrees(coordinates[0], coordinates[1])
+            radius = Angle.fromDegrees(coordinates[2])
```
andy-slac, line 137: I'm wondering now whether at least `lat` for center could be specified as `+/-Inf`. The standard is terribly underspecified.

python/lsst/sphgeom/_continue_class.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+@_continueClass
+class ConvexPolygon:  # noqa: F811
+    """A rectangle in spherical coordinate space that contains its boundary."""
+
+    def to_ivoa_pos(self) -> str:
+        # Docstring inherited.
+        vertices = self.getVertices()
+        coords = [LonLat(v) for v in vertices]
```
andy-slac, line 215: Could avoid making a list here, iterator should be OK?

andy-slac (approved):
Looks great, a couple of minor comments.

python/lsst/sphgeom/_continue_class.py
timj: They don't seem to have any examples that use Inf anywhere in CIRCLE so I will ignore for now.

PR #66, opened 2024-05-22
DM-44487: Fix sphgeom PyPi upload

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "lsst-sphgeom"
 description = "A spherical geometry library."
+requires-python = ">=3.8.0"
 license = {text = "BSD 3-Clause License"}
 readme = "README.md"
 authors = [
@@ -25,6 +26,8 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
     "Topic :: Scientific/Engineering :: Astronomy",
 ]
 keywords = ["lsst"]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -62,6 +62,7 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
+          name: sphgeom-sdist
           path: dist/*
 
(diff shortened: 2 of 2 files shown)


timj:
What's the motivation for dropping 3.8 and 3.9? I know we don't care that much but did they break somehow?

pyproject.toml
```diff
@@ -22,9 +22,9 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
     "Operating System :: MacOS :: MacOS X",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
timj, line 27: Might want to add:

```toml
requires-python = ">=3.10.0"
```

in the toml section as well.

mwittgen:
Build time and we don't  support it any more on other packages.

timj (approved):
Thanks. My main question is the python version change.

timj:
> Build time and we don't support it any more on other packages.

Okay. I don't think anyone is going to care.

fritzm:
Well, you might reasonably think that, but we might have Qserv bustage...   We're still on 3.8 for Python integration, regrettably, where Nate Pease left it...

mwittgen:
OK, putting 3.8 back.

fritzm:
Oh, but nevermind -- recalling we pull in `sphgeom` as a submodule for precisely these reasons :-)  So no problem.

mwittgen:
@fritzm Let's leave 3.8/9 in for now. Better do topical PRs and do this later.

fritzm (approved):
LGTM, thanks!

timj:
> Oh, but nevermind -- recalling we pull in sphgeom as a submodule for precisely these reasons :-) So no problem.

@fritzm it would be a problem, because no-one would know when we broke 3.8 if we were never testing it in the github action. Note that I was completely assuming that at any time I could go straight to 3.11 or whatever the current pipeline standard was. I've only been keeping it running with 3.8 because nothing was breaking with 3.8.

fritzm:
Right.  What I meant to say was "no problem for Qserv" (we'll continue to build successfully against our effectively "pinned" github SHA in the Qserv sphgeom submodule just fine until we actively choose to port forward to latest.)

PR #72, opened 2024-10-04
DM-46239:  Add pixelization envelopes for Intersection/UnionRegion

mwittgen:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -88,7 +88,6 @@ public:
     }
     static std::unique_ptr<CompoundRegion> decode(std::uint8_t const *buffer, size_t n);
     ///@}
-    static void flatten(Region const &region, std::vector<std::unique_ptr<Region>> &result);
 
 protected:
 
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -88,6 +88,7 @@ public:
     }
     static std::unique_ptr<CompoundRegion> decode(std::uint8_t const *buffer, size_t n);
     ///@}
+    static void flatten(Region const &region, std::vector<std::unique_ptr<Region>> &result);
 
 protected:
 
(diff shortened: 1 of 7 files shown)


include/lsst/sphgeom/CompoundRegion.h
```diff
@@ -88,6 +88,7 @@ class CompoundRegion : public Region {
     }
     static std::unique_ptr<CompoundRegion> decode(std::uint8_t const *buffer, size_t n);
     ///@}
+    static void flatten(Region const &region, std::vector<std::unique_ptr<Region>> &result);
```
erykoff, line 91: I don't think this should be here?

erykoff (commented):
First: a typo and a leftover flatten.  Second: I'm worried that findPixels is recursive, I thought that there was discussion on the ticket that this is bad?  (This may be way out of my wheelhouse though).

include/lsst/sphgeom/CompoundRegion.h
mwittgen: nope

mwittgen:
Allowing to loop over a flattened structures, we wanted to push to another ticket when it's clearer how to define a better API.

src/PixelFinder.h
```diff
@@ -156,23 +157,34 @@ template <
...
-            s, dynamic_cast<ConvexPolygon const &>(r), level, maxRanges);
+                s, *polygon, level, maxRanges);
         find();
+    } else if (auto union_region = dynamic_cast<UnionRegion const *>(&r)) {
+        Region const &region1 = union_region->getOperand(0);
+        Region const &region2 = union_region->getOperand(1);
+        auto rs1 = findPixels<Finder, InteriorOnly>(region1, maxRanges, level);
+        auto rs2 = findPixels<Finder, InteriorOnly>(region2, maxRanges, level);
+        s = rs1.join(rs2);
+    }else if (auto intersection_region = dynamic_cast<IntersectionRegion const *>(&r)) {
```
TallJimbo, line 180:
```suggestion
    } else if (auto intersection_region = dynamic_cast<IntersectionRegion const *>(&r)) {
```

TallJimbo (approved):
Looks good.  I wish we had an alternative to a `getRegions` that deep-copies everything just to let Python iterate over it, but given pybind11's inflexibility re `unique_ptr` vs. `shared_ptr` holders for the same type, I imagine we're pretty much out of options.

TallJimbo:
I think we're fine with letting this continue to be a fully recursive implementation and punt flattening that out to a new ticket.

PR #74, opened 2024-11-07
DM-46248: Allow CompoundRegion to be composed of many regions

andy-slac:
To avoid very deep recursion when CompoundRegion includes many regions it is switched from 2-Region array to a non-empty vector of Regions. Python wrapper should be backward compatible for constructing the instance but it now takes arbitrary number of positional arguments. New method `nOperands()` was added to access the actual number of regions. (Python interface could be made more natural by adding iteration over regions).

There is still a performance issue with `relate()` implementation which has to iterate over all operands almost always just because we calculate all types of relations in the same method. To help performance when we only need to know that twio regions are disjoint, new `Region.isDisjoint` method is added.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -36,7 +36,7 @@
 
 #include <iosfwd>
 #include <iterator>
-#include <array>
+#include <vector>
 #include <cstdint>
 
 #include "Region.h"
@@ -55,9 +55,8 @@ class Ellipse;
 class CompoundRegion : public Region {
 public:
     //@{
-    /// Construct by copying or taking ownership of operands.
-    CompoundRegion(Region const &first, Region const &second);
-    explicit CompoundRegion(std::array<std::unique_ptr<Region>, 2> operands) noexcept;
+    /// Construct by taking ownership of operands.
+    explicit CompoundRegion(std::vector<std::unique_ptr<Region>> operands) noexcept;
     //@}
 
     CompoundRegion(CompoundRegion const &);
@@ -67,6 +66,9 @@ public:
     // to guarantee memory safety for operand accessors in Python.
     CompoundRegion &operator=(CompoundRegion const &) = delete;
     CompoundRegion &operator=(CompoundRegion &&) = delete;
+
+    // Return number of operands
+    size_t nOperands() const { return _operands.size(); }
 
     // Return references to the operands.
     Region const & getOperand(std::size_t n) const {
@@ -95,11 +97,14 @@ protected:
     std::vector<std::uint8_t> _encode(std::uint8_t tc) const;
 
     // Implementation helper for decode().
-    static std::array<std::unique_ptr<Region>, 2> _decode(
+    static std::vector<std::unique_ptr<Region>> _decode(
         std::uint8_t tc, std::uint8_t const *buffer, std::size_t nBytes);
 
+    // Provide read access to all operands for quick iteration.
+    std::vector<std::unique_ptr<Region>> const& operands() const { return _operands; }
+
 private:
-    std::array<std::unique_ptr<Region>, 2> _operands;
+    std::vector<std::unique_ptr<Region>> _operands;
 };
 
 /// UnionRegion is a lazy point-set union of its operands.
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -103,6 +103,10 @@ protected:
     // Provide read access to all operands for quick iteration.
     std::vector<std::unique_ptr<Region>> const& operands() const { return _operands; }
 
+    // Flatten vector of regions in-place.
+    template <typename Compound>
+    void flatten_operands();
+
 private:
     std::vector<std::unique_ptr<Region>> _operands;
 };
@@ -115,7 +119,8 @@ class UnionRegion : public CompoundRegion {
 public:
     static constexpr std::uint8_t TYPE_CODE = 'u';
 
-    using CompoundRegion::CompoundRegion;
+    /// Construct by taking ownership of operands.
+    explicit UnionRegion(std::vector<std::unique_ptr<Region>> operands);
 
     // Region interface.
     std::unique_ptr<Region> clone() const override { return std::make_unique<UnionRegion>(*this); }
@@ -154,7 +159,8 @@ class IntersectionRegion : public CompoundRegion {
 public:
     static constexpr std::uint8_t TYPE_CODE = 'i';
 
-    using CompoundRegion::CompoundRegion;
+    /// Construct by taking ownership of operands.
+    explicit IntersectionRegion(std::vector<std::unique_ptr<Region>> operands);
 
     // Region interface.
     std::unique_ptr<Region> clone() const override { return std::make_unique<IntersectionRegion>(*this); }
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -119,12 +119,18 @@ public:
 
     // Region interface.
     std::unique_ptr<Region> clone() const override { return std::make_unique<UnionRegion>(*this); }
+    bool isEmpty() const override;
     Box getBoundingBox() const override;
     Box3d getBoundingBox3d() const override;
     Circle getBoundingCircle() const override;
     using Region::contains;
     bool contains(UnitVector3d const &v) const override;
     Relationship relate(Region const &r) const override;
+    TriState overlaps(Region const& other) const override;
+    TriState overlaps(Box const &) const override;
+    TriState overlaps(Circle const &) const override;
+    TriState overlaps(ConvexPolygon const &) const override;
+    TriState overlaps(Ellipse const &) const override;
     std::vector<std::uint8_t> encode() const override { return _encode(TYPE_CODE); }
 
     ///@{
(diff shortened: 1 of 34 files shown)


andy-slac:
@TallJimbo, could you look at my changes to `sphgeom`, there are two changes which should help with the performance of `UnionRegion`:
- `CompoundRegion` is now a non-empty vector of `Region`s
- New method `Region.isDisjoint` to calculate only that type of region relation to avoid using more expensive `relate`. Maybe we need a better name for the new method, is there is already `isDisjointFrom` in some specific subclasses.

include/lsst/sphgeom/CompoundRegion.h
```diff
@@ -125,6 +125,7 @@ class UnionRegion : public CompoundRegion {
     using Region::contains;
     bool contains(UnitVector3d const &v) const override;
     Relationship relate(Region const &r) const override;
+    bool isDisjoint(Region const& other) const override;
```
TallJimbo, line 128: Given the precedent, I do think `isDisjointFrom` would be a slightly better name.  But if that leads to lots of complications with overloading and shadowing it may not be worth it.

Or, if the exiting `isDisjointFrom` implementations are all exact rather than conservative, maybe we call this one `isDefinitelyDisjointFrom`?  That's verbose, but it might make it a little more clear that `False` only means "maybe disjoint".

include/lsst/sphgeom/Region.h
```diff
@@ -136,6 +136,25 @@ class Region {
     virtual Relationship relate(Ellipse const &) const = 0;
     ///@}
 
+    ///@{
+    /// `disjoint` tests whether this region is disjoint from other region.
+    /// This method returns a subset of information returned from `relate`,
+    /// the reason for its existence is that in same cases it may be
+    /// implemented more efficiently.
+    ///
+    /// The meaning of the returned value is the same as for `relate` metod -
```
TallJimbo, line 145:
```suggestion
    /// The meaning of the returned value is the same as for `relate` method -
```

include/lsst/sphgeom/CompoundRegion.h
```diff
@@ -68,6 +67,9 @@ class CompoundRegion : public Region {
     CompoundRegion &operator=(CompoundRegion const &) = delete;
     CompoundRegion &operator=(CompoundRegion &&) = delete;
 
+    // Return number of operands
+    unsigned int nOperands() const { return _operands.size(); }
```
TallJimbo, line 71: Unless there's precedent in sphgeom for using `unsigned int` for sizes of things, I think `std::size_t` would be better.

src/CompoundRegion.cc
```diff
@@ -61,28 +61,38 @@ std::uint64_t consumeDecodeU64(std::uint8_t const *&buffer, std::uint8_t const *
...
-
-CompoundRegion::CompoundRegion(
-    std::array<std::unique_ptr<Region>, 2> operands) noexcept
-        : _operands(std::move(operands)) {}
+CompoundRegion::CompoundRegion(std::vector<std::unique_ptr<Region>> operands) noexcept
+        : _operands(std::move(operands))
+{
+    if (_operands.empty()) {
+        throw std::invalid_argument("CompoundRegion requires non-empty region list.");
+    }
```
TallJimbo, line 86: An alternative to banning empty lists that might be more convenient would be to declare:

- an empty `UnionRegion` is an empty region, i.e. contains no points;
- an empty `IntersectionRegion` is a full region, i.e. contains the full sphere.

I think that's the natural limiting behavior for all of the `isDisjoint` and `relate` implementations.

src/CompoundRegion.cc
```diff
@@ -151,21 +159,43 @@ Circle UnionRegion::getBoundingCircle() const {
...
-        | ((r1 & WITHIN) & (r2 & WITHIN))
-        // If either operand contains the given region, the union contains it.
-        | ((r1 & CONTAINS) | (r2 & CONTAINS));
+        if ((rel & WITHIN) != WITHIN) {
+            result &= ~WITHIN;
+        }
+        // If any operand contains the given region, the union contains it.
+        // NOTE: The logic here is incomplete there may be cases when union
+        // of regions can fully contain a region even though individual regions
+        // do not contain it.
```
TallJimbo, line 189: Thankfully this is still consistent with the carefully conservative `relate` documentation:
https://github.com/lsst/sphgeom/blob/781aacb6e4d47f4a80a54d607375a7f372c316b2/include/lsst/sphgeom/Region.h#L124-L127

src/CompoundRegion.cc
```diff
@@ -181,22 +211,45 @@ Circle IntersectionRegion::getBoundingCircle() const {
...
+        // NOTE: there may be cases when intersection is disjoint even though
+        // not all operands are disjoint.
+        if ((rel & DISJOINT) == DISJOINT) {
+            result |= DISJOINT;
+        }
+        // If any operand is within the given region, the intersection is
         // within it.
-        | ((r1 & WITHIN) | (r2 & WITHIN));
+        // NOTE: There may be cases when intersectiuon is within the region
+        // even though not all operand are within.
```
TallJimbo, line 253: (NOTEs here are also consistent with the `relate` documentation.)

python/lsst/sphgeom/_compoundRegion.cc
```diff
@@ -41,38 +41,60 @@ namespace sphgeom {
...
 
 template <>
 void defineClass(py::class_<CompoundRegion, std::unique_ptr<CompoundRegion>, Region> &cls) {
+    cls.def("nOperands", &CompoundRegion::nOperands);
     cls.def(
         "cloneOperand",
         [](CompoundRegion const &self, std::ptrdiff_t n) {
-            return self.getOperand(python::convertIndex(2, n)).clone();
+            int nOperands = self.nOperands();
+            return self.getOperand(python::convertIndex(nOperands, n)).clone();
```
TallJimbo, line 79: I'd love to have a way to let Python code iterate over regions without having to clone them, if we can do it safely.  As long as we don't provide a way to modify a compound region after it has been constructed, I believe pybind11's `return_value_policy::reference_internal` should make it safe to return a reference to an operand.

src/CompoundRegion.cc
```diff
@@ -61,28 +61,38 @@ std::uint64_t consumeDecodeU64(std::uint8_t const *&buffer, std::uint8_t const *
...
 
-CompoundRegion::CompoundRegion(Region const &first, Region const &second)
-        : _operands{first.clone(), second.clone()} {}
-
-CompoundRegion::CompoundRegion(
-    std::array<std::unique_ptr<Region>, 2> operands) noexcept
-        : _operands(std::move(operands)) {}
+CompoundRegion::CompoundRegion(std::vector<std::unique_ptr<Region>> operands) noexcept
+        : _operands(std::move(operands))
+{
```
TallJimbo, line 83: It might be worth it to have some logic here (or maybe in subclass constructors) to look for operands that are themselves compound regions of the same type, in order to flatten them out.

TallJimbo (approved):

include/lsst/sphgeom/CompoundRegion.h
andy-slac: Hmm, but `isDisjointFrom` is also "definitely disjoint", would `isDefinitelyDisjointFrom` make people think that `isDisjointFrom` is less definite?

src/CompoundRegion.cc
andy-slac: I actually started with that approach but it requires extra logic in few places which was not quite obvious to me, and I was not sure that it was useful at all. I'll try to restore that and add few unit tests.

src/CompoundRegion.cc
andy-slac: Indeed, I read `relate` documentation later, but forgot to remove my dummy comments.

src/CompoundRegion.cc
andy-slac: That logic would also need to handle nested unions and intersections differently, I guess?

TallJimbo: Yes, but only in the sense that we don't try to simplify a compound region of the opposite type.  I don't think we need to go looking for intersections buried under unions inside an intersection, for instance; if we did I think it'd be very easy to do more harm than good.

include/lsst/sphgeom/CompoundRegion.h
andy-slac: Another option - call this new method `overlaps(other)` but return tri-state value from it (yes/no/maybe or `bool | None` on Python side). That way we can avoid confusion with existing methods and also make `maybe` distinguishable from two other answers.

TallJimbo: I like that idea, though I suspect it might be a lot of work (without much payoff) to make it actually return `true` sometimes.

andy-slac: In those cases it's probably OK to return `maybe`, just like current code is probably doing for `DISTICT` in `relate`. C++ does not have a standard class for tri-state, Boost has `tristate` but I guess we do not want to add it as dependency. Maybe simple `enum` should be OK on C++ side.

python/lsst/sphgeom/_compoundRegion.cc
andy-slac: I have added `__len__` and `__iter__` to CompoundRegion using `reference_internal` for latter. I did not make `_operands` cost, as on another commit I added method to flatten it in place after construction.

include/lsst/sphgeom/CompoundRegion.h
andy-slac: I have replaced it with `overlaps()` which returns an instance of a new `TriState` class. IntersectRegion implementation for that is not very useful as it cannot correctly guess "definitely overlaps" case, but I think we now mostly care about unions.

PR #75, opened 2024-12-10
DM-47947: add functions for processing concatenations of encoded regions

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -36,7 +36,7 @@
 
 #include <iosfwd>
 #include <iterator>
-#include <array>
+#include <vector>
 #include <cstdint>
 
 #include "Region.h"
@@ -55,9 +55,8 @@ class Ellipse;
 class CompoundRegion : public Region {
 public:
     //@{
-    /// Construct by copying or taking ownership of operands.
-    CompoundRegion(Region const &first, Region const &second);
-    explicit CompoundRegion(std::array<std::unique_ptr<Region>, 2> operands) noexcept;
+    /// Construct by taking ownership of operands.
+    explicit CompoundRegion(std::vector<std::unique_ptr<Region>> operands) noexcept;
     //@}
 
     CompoundRegion(CompoundRegion const &);
@@ -67,6 +66,9 @@ public:
     // to guarantee memory safety for operand accessors in Python.
     CompoundRegion &operator=(CompoundRegion const &) = delete;
     CompoundRegion &operator=(CompoundRegion &&) = delete;
+
+    // Return number of operands
+    size_t nOperands() const { return _operands.size(); }
 
     // Return references to the operands.
     Region const & getOperand(std::size_t n) const {
@@ -95,11 +97,14 @@ protected:
     std::vector<std::uint8_t> _encode(std::uint8_t tc) const;
 
     // Implementation helper for decode().
-    static std::array<std::unique_ptr<Region>, 2> _decode(
+    static std::vector<std::unique_ptr<Region>> _decode(
         std::uint8_t tc, std::uint8_t const *buffer, std::size_t nBytes);
 
+    // Provide read access to all operands for quick iteration.
+    std::vector<std::unique_ptr<Region>> const& operands() const { return _operands; }
+
 private:
-    std::array<std::unique_ptr<Region>, 2> _operands;
+    std::vector<std::unique_ptr<Region>> _operands;
 };
 
 /// UnionRegion is a lazy point-set union of its operands.
--- a/include/lsst/sphgeom/CompoundRegion.h
+++ b/include/lsst/sphgeom/CompoundRegion.h
@@ -103,6 +103,10 @@ protected:
     // Provide read access to all operands for quick iteration.
     std::vector<std::unique_ptr<Region>> const& operands() const { return _operands; }
 
+    // Flatten vector of regions in-place.
+    template <typename Compound>
+    void flatten_operands();
+
 private:
     std::vector<std::unique_ptr<Region>> _operands;
 };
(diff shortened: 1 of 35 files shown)


include/lsst/sphgeom/CompoundRegion.h
```diff
@@ -95,11 +97,21 @@ class CompoundRegion : public Region {
...
     // Implementation helper for decode().
-    static std::array<std::unique_ptr<Region>, 2> _decode(
+    static std::vector<std::unique_ptr<Region>> _decode(
         std::uint8_t tc, std::uint8_t const *buffer, std::size_t nBytes);
 
+    // Provide read access to all operands for quick iteration.
+    std::vector<std::unique_ptr<Region>> const& operands() const { return _operands; }
+
+    // Check if we have any operands.
+    bool empty() const { return _operands.empty(); }
```
TallJimbo, line 107: Given that `Region` also has an `isEmpty` method that's about whether there are any points in the region, I think this might be more confusing than helpful.

src/CompoundRegion.cc
```diff
@@ -61,28 +62,53 @@ std::uint64_t consumeDecodeU64(std::uint8_t const *&buffer, std::uint8_t const *
 template <typename F>
 auto getUnionBounds(UnionRegion const &compound, F func) {
     auto bounds = func(compound.getOperand(0));
-    bounds.expandTo(func(compound.getOperand(1)));
+    for (unsigned i = 1; i < compound.nOperands(); ++ i) {
```
TallJimbo, line 65:
```suggestion
    for (std::size_t i = 1; i < compound.nOperands(); ++ i) {
```

src/CompoundRegion.cc
```diff
@@ -61,28 +62,53 @@ std::uint64_t consumeDecodeU64(std::uint8_t const *&buffer, std::uint8_t const *
...
+        bounds.expandTo(func(compound.getOperand(i)));
+    }
     return bounds;
 }
 
 template <typename F>
 auto getIntersectionBounds(IntersectionRegion const &compound, F func) {
     auto bounds = func(compound.getOperand(0));
-    bounds.clipTo(func(compound.getOperand(1)));
+    for (unsigned i = 1; i < compound.nOperands(); ++ i) {
```
TallJimbo, line 74:
```suggestion
    for (std::size_t i = 1; i < compound.nOperands(); ++ i) {
```

include/lsst/sphgeom/CompoundRegion.h
andy-slac: OK, I'll drop it, I can use `nOperands()` which is public and will likely produce the same code.

src/Region.cc
```diff
@@ -73,15 +90,63 @@ std::unique_ptr<Region> Region::decode(std::uint8_t const * buffer, size_t n) {
...
+            region_begin = region_end;
+            ++region_begin;
+            region_end = std::find(region_begin, s.end(), ':');
+        }
+        auto bytes = base64::decode_into<std::vector<std::uint8_t>>(region_begin, region_end);
+        union_args.push_back(decode(bytes));
+        return std::unique_ptr<UnionRegion>(new UnionRegion(std::move(union_args)));
+    } else {
+        auto bytes = base64::decode_into<std::vector<std::uint8_t>>(region_begin, region_end);
+        return decode(bytes);
```
andy-slac, line 110: My guess is that in this scheme an empty UnionRegion is represented by an empty string? I'm not sure that it could happen in our daf_butler queries, but it may be worth to support it in general, and this code will probably throw an exception for an empty string?

tests/test_CompoundRegion.py
```diff
@@ -228,6 +229,18 @@ def testBounding(self):
...
+        union regions.
+        """
+        # Test with the full UnionRegion encoded, then base64-encoded.
+        s1 = b64encode(self.instance.encode()).decode("ascii")
+        self.assertCompoundRegionsEqual(Region.decodeBase64(s1), self.instance)
+        # Test alternate form with union members concatenated with ':' after
+        # base64-encoding.
+        s2 = ":".join(b64encode(region.encode()).decode("ascii") for region in self.instance)
+        self.assertCompoundRegionsEqual(Region.decodeBase64(s2), self.instance)
+
```
andy-slac, line 243: Maybe add a test with an empty UnionRegion?

include/lsst/sphgeom/Region.h
```diff
@@ -149,9 +166,55 @@ class Region {
     static std::unique_ptr<Region> decode(std::uint8_t const * buffer, size_t n);
     ///@}
 
+    ///@{
+    /// `decodeBase64` deserializes a Region from an ASCII string produced by
+    /// encode and then base64-encoding that result.
+    ///
+    /// This method also interprets ':' as a delimiter for the elements of a
+    /// UnionRegion, to support cases where a union of region is constructed
+    /// server-side in a database as an concatenation with that delimiter.
```
andy-slac, line 175: Would it be useful to add support for IntersectionRegion as well, via a different delimiter?

tests/test_CompoundRegion.py
```diff
@@ -321,6 +321,33 @@ def testBounding(self):
...
+        self.assertEqual(run_overlaps([(self.box, self.circle)]), True)
+        self.assertEqual(run_overlaps([(self.box, self.faraway)]), False)
+        self.assertEqual(run_overlaps([(self.circle, self.faraway)]), False)
+        self.assertEqual(run_overlaps([(self.instance, self.box)]), None)
+        self.assertEqual(run_overlaps([(self.box, self.circle), (self.box, self.faraway)]), True)
+        self.assertEqual(run_overlaps([(self.faraway, self.circle), (self.box, self.faraway)]), False)
+        self.assertEqual(run_overlaps([(self.instance, self.box), (self.circle, self.faraway)]), None)
+        self.assertEqual(run_overlaps([(self.instance, self.box), (self.circle, self.box)]), True)
+        self.assertEqual(run_overlaps([(self.circle, self.box), (self.instance, self.box)]), True)
+
```
andy-slac, line 350: Maybe add an exceptional test without `&`?

andy-slac (approved):
Looks great, few minor comments.

include/lsst/sphgeom/Region.h
TallJimbo: I thought about that too, but I think it'd end up dead code.  And if I'm wrong, it might need to be more complex than allowing just one delimiter or the other (i.e. we'd need to work out associativity).

PR #80, opened 2025-01-04
DM-48290: Add hpp files to python sdist

mwittgen:
## Checklist

- [ ] ran Jenkins


roceb (approved):

timj:
Makes me think that we should be checking that the sdist actually builds before we do the upload.

timj (approved):

PR #105, opened 2026-04-17
DM-54630: Add GIL-free envelope_many function

dhirving:
Add a function for computing many calls to Pixelization::envelope(), with the Python GIL dropped.  This is intended to optimize the bulk load of Butler skypix overlap rows when inserting dimension records.

## Checklist

- [ ] ran Jenkins


dhirving:
I decided not to pursue this.  It does give a modest speedup (about 30%) with many concurrent multi-threaded Butler dimension record inserts, but the process is still heavily GIL-bound from other code and it's not worth breaking `daf_butler` compatibility with older `sphgeom`.

PR #106, opened 2026-05-26
DM-53569: Add Region to IVOA STC-S conversion

timj:
## Checklist

- [x] ran Jenkins
- [x] added a release note for user-visible changes to `doc/changes`


python/lsst/sphgeom/_continue_class.py
```diff
@@ -172,6 +205,40 @@ def to_ivoa_pos(self) -> str:
...
+        -----
+        See
+        http://www.ivoa.net/Documents/Notes/STC-S/20091030/NOTE-STC-S-1.33-20091030.html
+        for the format definition. Supported region types are ``Circle``,
+        ``Polygon``, ``Ellipse``, and the ``Union`` / ``Intersection``
+        compound operators. ``Box`` regions cannot be converted directly
+        because STC-S has no latitude-parallel range region.
+        """
+        head, _, tail = self._ivoa_stcs_body().partition(" ")
+        return f"{head} {frame} {tail}"
```
andy-slac: It could be more efficient to pass `frame` to `_ivoa_stcs_body()` so that string formatting can be done once only instead of format/partition/re-format? I imagine this would be called on very many regions so performance may be important.

python/lsst/sphgeom/_continue_class.py
```diff
@@ -215,3 +298,44 @@ def to_ivoa_pos(self) -> str:
...
+        lat = center.getLat().asDegrees()
+        alpha = self.getAlpha().asDegrees()
+        beta = self.getBeta().asDegrees()
+        pa = _ellipse_position_angle_degrees(self)
+        return f"Ellipse {lon} {lat} {alpha} {beta} {pa}"
+
+
+@_continueClass
+class UnionRegion:  # noqa: F811
+    """A union of two regions on the unit sphere."""
```
andy-slac: I think it's an arbitrary number of regions, not just two, same for `IntersectionRegion`.

python/lsst/sphgeom/_continue_class.py
```diff
@@ -215,3 +298,44 @@ def to_ivoa_pos(self) -> str:
...
+
+    def _ivoa_stcs_body(self) -> str:
+        # Docstring inherited.
+        operands = [self.cloneOperand(i)._ivoa_stcs_body() for i in range(self.nOperands())]
```
andy-slac: Do you really need `cloneOperand`? I think you can just iterate over `self` to get each operand.

andy-slac (approved):
Looks good, a couple of minor comments.

python/lsst/sphgeom/_continue_class.py
timj: Looks like a 5% gain by moving the frame lower down.

python/lsst/sphgeom/_continue_class.py
timj: Switching to `self` slows things down by 30% so I think we leave it.

andy-slac: Hmm, that's surprising, I'd think that avoiding clone of each region should make things faster. 🤔

timj: I think everything is so quick in general that using the python iterator on self adds extra overhead.

python/lsst/sphgeom/_continue_class.py
timj: I do apologize. Claude failed to follow instructions and made comments on this PR in my name. I will delete them.

timj: Anyhow, since you were concerned about performance I got claude to convert everything to C++ and gained an extra 20% performance for Polygon (the only one that matters) and massive speed up for Circle in case it matters.

timj:
@andy-slac since everything was rewritten (but is the same logic) in C++ let me know if you feel like doing a new review.

andy-slac:
@timj, yes, I'll check it.

include/lsst/sphgeom/Region.h
```diff
@@ -198,6 +214,32 @@ class Region {
...
+
+    /// `toIvoaStcsBody` returns the STC-S body of this region. When `frame`
+    /// is non-empty it is inserted immediately after the shape keyword; an
+    /// empty `frame` produces the bare body, which compound regions use
+    /// when recursing into operands so the frame keyword appears only at
+    /// the outermost level.
+    ///
+    /// The default implementation throws `IvoaStcsNotImplemented`; concrete
+    /// subclasses override it.
+    virtual std::string toIvoaStcsBody(std::string const & frame = "") const {
```
andy-slac: Do we want this method to be public, this feels like an implementation detail?

andy-slac (approved):
Looks good.

include/lsst/sphgeom/Region.h
timj: Yes. Fixed. It was left over from the earlier approach that only had Polygon in C++.

PR #112, opened 2026-08-26
DM-55954: Fix to_chars on macOS by reimplementing

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/testFloatFormat.cc
+++ b/tests/testFloatFormat.cc
@@ -1,0 +1,224 @@
+/*
+ * This file is part of sphgeom.
+ *
+ * Developed for the LSST Data Management System.
+ * This product includes software developed by the LSST Project
+ * (http://www.lsst.org).
+ * See the COPYRIGHT file at the top-level directory of this distribution
+ * for details of code ownership.
+ *
+ * This software is dual licensed under the GNU General Public License and also
+ * under a 3-clause BSD license. Recipients may choose which of these licenses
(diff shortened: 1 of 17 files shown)


tests/testFloatFormat.cc
```diff
@@ -0,0 +1,166 @@
...
+int minimalDigits(double v) {
+    char buf[32];
+    for (int n = 1; n < 17; ++n) {
+        std::snprintf(buf, sizeof(buf), "%.*e", n - 1, v);
+        if (std::strtod(buf, nullptr) == v) {
+            return n;
+        }
+    }
+    return 17;
+}
```
andy-slac, line 81: With linear search and arbitrary inputs it may be more efficient to start with the longest string and try to shorten it until it does not round-trip. Probably not important for tests.

andy-slac:
I'm concerned that we are pessimising this for all platforms even though the problem exists only on a single platform. Is there a compile-time check that can tell that `std::to_chars` can or cannot work for doubles? How can we tell when it's OK to drop this workaround? And for a platform where `std::to_chars` that does not do doubles, do we care if representation is not the shortest one?

andy-slac (approved):

tests/testFloatFormat.cc
timj: Thanks. This made the tests go much faster.

PR #114, opened 2026-09-04
DM-56017: Add Python type annotations to sphgeom

andy-slac:
## Checklist

- [ ] ran Jenkins


python/lsst/sphgeom/__init__.pyi
```diff
@@ -0,0 +1,1110 @@
...
+    def getHeight(self) -> Angle: ...
+    @overload
+    def contains(self, unitVector: UnitVector3d) -> bool: ...
+    @overload
+    def contains(
+        self,
+        x: typing.Annotated[numpy.typing.ArrayLike, numpy.float64],
+        y: typing.Annotated[numpy.typing.ArrayLike, numpy.float64],
+        z: typing.Annotated[numpy.typing.ArrayLike, numpy.float64],
+    ) -> typing.Annotated[numpy.typing.ArrayLike, numpy.bool]: ...
```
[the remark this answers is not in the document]
andy-slac: Not sure that this is `numpy.True_` 🙂 Looks like there was a history of numpy trying to get rid of `numpy.bool` but it got reverted. Current docs say `numpy.bool_` is an alias to `numpy.bool` but I'll change `bool` to `bool_` to keep my copilot happy.

python/lsst/sphgeom/_healpixPixelization.py
```diff
@@ -151,16 +151,12 @@ def _interior_pixels_from_region(self, nside: int, region: Region):
...
-            retry = False
             try:
                 pixels = hpg.query_polygon_vec(nside, vertices, inclusive=True)
             except RuntimeError:
                 # This happens if the polygon is determined to have a
                 # degenerate corner (which can happen with rounding).
                 # In this case, redo with a bounding circle.
-                retry = True
-
-            if retry:
```
timj, line 163: Are you sure this wasn't deliberate so that you would get a cleaner exception?

python/lsst/sphgeom/_continue_class.py
```diff
@@ -214,12 +216,12 @@ def to_ivoa_pos(self) -> str:
...
 
 @_continueClass
-class ConvexPolygon:  # noqa: F811
+class ConvexPolygon:  # type: ignore[no-redef] # noqa: F811
     """A rectangle in spherical coordinate space that contains its boundary."""
 
     def to_ivoa_pos(self) -> str:
         # Docstring inherited.
-        coords = (LonLat(v) for v in self.getVertices())
+        coords = (LonLat(v) for v in self.getVertices())  # type: ignore[attr-defined]
```
timj, line 224: why can't it work out the type for this?

timj (approved):
Looks good. Hopefully at some point we can move some of these to the pybind11 code itself.

python/lsst/sphgeom/_healpixPixelization.py
andy-slac: Hmm, I did not think about that. I'll revert this patch.

python/lsst/sphgeom/_continue_class.py
andy-slac: `mypy` can't figure out `@_continueClass` magic, it complains that `self` does not have `getVertices` attribute.

PR #115, opened 2026-09-22
DM-56189: Add contains(Region) to the Region base class

timj:
Only ConvexPolygon provided region containment, so code holding a region through a Region handle had to fall back to relate(). The new method is implemented in terms of relate() and shares its conservative semantics: a true result is conclusive, a false one only means containment could not be established.

Box and Circle rewrap the overload explicitly because pybind11 lets a subclass's own contains overloads hide the ones bound on the base class.

## Checklist

- [x] ran Jenkins
- [x] added a release note for user-visible changes to `doc/changes`
- [ ] updated mypy stubs in `python/lsst/sphgeom/__init__.pyi` for any Python API changes


include/lsst/sphgeom/Region.h
```diff
@@ -132,6 +132,21 @@ class Region {
...
+    /// semantics: a true return value is conclusive, but a false one only
+    /// means that containment could not be established. How readily that
+    /// happens depends on the dynamic types of both regions, since `relate`
+    /// may substitute a simplified bounding region for either side. `Ellipse`
+    /// relates via its bounding circle and so never reports containment at
+    /// all, even of itself.
+    ///
+    /// Call `relate` directly when "does not contain" has to be
+    /// distinguished from "could not tell".
+    bool contains(Region const &) const;
```
andy-slac: Should this overload (and maybe two others above) be virtual?

python/lsst/sphgeom/_box.cc
```diff
@@ -102,6 +102,7 @@ void defineClass(py::classh<Box, Region> &cls) {
     cls.def("contains", (bool (Box::*)(LonLat const &) const) & Box::contains);
     cls.def("contains", (bool (Box::*)(Box const &) const) & Box::contains);
     // Rewrap these base class methods since there are overloads in this subclass
+    cls.def("contains", (bool (Region::*)(Region const &) const) & Region::contains, "region"_a);
```
andy-slac:
```suggestion
    cls.def("contains", (bool (Box::*)(Region const &) const) & Box::contains, "region"_a);
```
like the wrapper below?

python/lsst/sphgeom/_circle.cc
```diff
@@ -82,6 +82,7 @@ void defineClass(py::classh<Circle, Region> &cls) {
     cls.def("contains",
             (bool (Circle::*)(Circle const &) const) & Circle::contains);
     // Rewrap these base class methods since there are overloads in this subclass
+    cls.def("contains", (bool (Region::*)(Region const &) const) & Region::contains, "region"_a);
```
andy-slac:
```suggestion
    cls.def("contains", (bool (Circle::*)(Region const &) const) & Circle::contains, "region"_a);
```
?

andy-slac (approved):
Looks good, but I think that new method should be virtual?

